PR #3, opened 2015-06-07
LibOVRIntegration

Squareys:
Hi @mosra, Hi everybody,

Since this definitely will be a thing now, let's move discussion of LibOVRIntegration to here. Previous discussion can be found in https://github.com/mosra/magnum-examples/pull/10.

If you are okay with it, only the most important compositing layer (ovrLayerEyeFov) will be wrapped for now and I will create an issue which lists the missing wrappers and implement them when needed.

Current TODOs:
- [x] Finish wrapping compositor.
- [x] Cleanup code.
  - [x] Cleanup code formatting. 
  - [x] Change `// comments` to `/* comments */`.
  - [x] Change `MagnumLibOvrIntegration_EXPORTS` to be all uppercase.
  - [x] Remove `#endif` description comments.
  - [x] Add copyrights properly.
  - [x] Fix posef test.
  - [x] Use `#include<memory>`.
  - [x] Add enumset operators.
  - [x] Fix `ovrSwapTextureSet* getOvrSwapTextureSet()` being a raw pointer.
  - [x] Change case of `MAGNUM_LibOvrIntegration_EXPORT`.
  - [x] Refactor context.
  - [x] Use absolute includes.
- [x] Rebase to cleanup commit history.
   - [x] Squash fixups.
- [x] Finish doc++.
- [x] Finish `FindOVR.cmake`.

### DIFF
--- a/modules/FindMagnumIntegration.cmake
+++ b/modules/FindMagnumIntegration.cmake
@@ -6,6 +6,7 @@
 #  MAGNUMINTEGRATION_FOUND      - Whether the library was found
 # This command alone is useless without specifying the components:
 #  Bullet                       - Bullet Physics integration library
+#  LibOvr                       - LibOVR integration library
 # Example usage with specifying additional components is:
 #  find_package(MagnumIntegration [REQUIRED|COMPONENTS]
 #               Bullet)
@@ -113,6 +114,15 @@ foreach(component ${MagnumIntegration_FIND_COMPONENTS})
             set(_MAGNUM_${_COMPONENT}INTEGRATION_LIBRARIES ${BULLET_LIBRARIES})
             set(_MAGNUM_${_COMPONENT}INTEGRATION_INCLUDE_PATH_NAMES MotionState.h)
             set(_MAGNUM_${_COMPONENT}INTEGRATION_INCLUDE_DIRS ${BULLET_INCLUDE_DIRS})
+        else()
+            unset(MAGNUM_${_COMPONENT}INTEGRATION_LIBRARY)
+        endif()
+    elseif(${component} STREQUAL LibOvr)
+        find_package(OVR)
+        if(OVR_FOUND)
+            set(_MAGNUM_${_COMPONENT}INTEGRATION_LIBRARIES ${OVR_LIBRARIES})
+            set(_MAGNUM_${_COMPONENT}INTEGRATION_INCLUDE_PATH_NAMES LibOvrIntegration.h)
+            set(_MAGNUM_${_COMPONENT}INTEGRATION_INCLUDE_DIRS ${PVR_INCLUDE_DIRS})
         else()
             unset(MAGNUM_${_COMPONENT}INTEGRATION_LIBRARY)
         endif()
--- a/modules/FindOVR.cmake
+++ b/modules/FindOVR.cmake
@@ -1,0 +1,62 @@
+# - Find OVR
+#
+# This module defines:
+#
+#  OVR_FOUND                - True if OVR library is found
+#  OVR_LIBRARY              - OVR dynamic library
+#  OVR_INCLUDE_DIR          - Include dir
+#
+
+#
+#   This file is part of Magnum.
+#
+#   Copyright © 2010, 2011, 2012, 2013, 2014, 2015
+#             Vladimír Vondruš <mosra@centrum.cz>
+#   Copyright © 2015
+#             Jonathan Hale <squareys@googlemail.com>
+#
+#   Permission is hereby granted, free of charge, to any person obtaining a
+#   copy of this software and associated documentation files (the "Software"),
+#   to deal in the Software without restriction, including without limitation
+#   the rights to use, copy, modify, merge, publish, distribute, sublicense,
+#   and/or sell copies of the Software, and to permit persons to whom the
+#   Software is furnished to do so, subject to the following conditions:
+#
+#   The above copyright notice and this permission notice shall be included
+#   in all copies or substantial portions of the Software.
+#
+#   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+#   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+#   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+#   THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#   DEALINGS IN THE SOFTWARE.
+#
+
+set(LIBOVR_ROOT ${CMAKE_INSTALL_PREFIX})
+
+if(NOT OVR_SDK_ROOT)
+    message(WARNING "OVR_SDK_ROOT is not set. Will try to find headers and library in ${CMAKE_INSTALL_PREFIX}." )
+else()
+    set(LIBOVR_ROOT ${OVR_SDK_ROOT}/LibOVR)
+endif()
+
+# find include directory
+find_path(OVR_INCLUDE_DIR NAMES OVR_CAPI.h HINTS ${LIBOVR_ROOT}/Include)
+
+if(WIN32)
+    if(MSVC)
+        find_library(OVR_LIBRARY NAMES LibOVR HINTS ${LIBOVR_ROOT}/Lib/Windows)
+    elseif(MINGW)
+        # linking against the MSVC dll with MinGW does not work directly. Instead, you need to
+        # link against a specific version. This will cause problems with newer oculus runtimes,
+        # though. (FIXME!)
+        # The clean way to link against libOVR with mingw would be to compile the linux version,
+        # 0.6.0.0-beta is not available for Linux, though.
+        find_library(OVR_LIBRARY NAMES LibOVRRT32_0_6.dll HINTS "C:/Windows/SysWOW64")
+    endif()
+endif()
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(OVR OVR_LIBRARY OVR_INCLUDE_DIR)
--- a/src/Magnum/LibOvrIntegration/Compositor.cpp
+++ b/src/Magnum/LibOvrIntegration/Compositor.cpp
@@ -1,0 +1,122 @@
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015
+              Vladimír Vondruš <mosra@centrum.cz>
+    Copyright © 2015
+              Jonathan Hale <squareys@googlemail.com>
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+#include "Magnum/LibOVRIntegration/Compositor.h"
+
+#include "Magnum/LibOVRIntegration/Conversion.h"
+#include "Magnum/LibOVRIntegration/Hmd.h"
+
+#include <OVR_CAPI_GL.h>
+
+namespace Magnum { namespace LibOvrIntegration {
+
+Layer::Layer(const LayerType type): _layer(), _type(type) {
+    _layer.Header.Type = ovrLayerType(Corrade::Containers::EnumSet<LayerType>::UnderlyingType(_type));
+    _layer.Header.Flags = ovrLayerFlag_TextureOriginAtBottomLeft;
+}
+
+Layer& Layer::setEnabled(bool enabled) {
+    if(enabled) {
+        _layer.Header.Type = ovrLayerType(Corrade::Containers::EnumSet<LayerType>::UnderlyingType(_type));
+    } else {
+        _layer.Header.Type = ovrLayerType_Disabled;
+    }
+    return *this;
+}
+
+//----------------------------------------------------------------
+
+LayerEyeFov::LayerEyeFov(): Layer(LayerType::EyeFov) {
+}
+
+LayerEyeFov& LayerEyeFov::setColorTexture(const int eye, SwapTextureSet& textureSet) {
+    _layer.EyeFov.ColorTexture[eye] = &textureSet.getOvrSwapTextureSet();
+
+    return *this;
+}
+
+LayerEyeFov& LayerEyeFov::setViewport(const int eye, const Range2Di& viewport) {
+    _layer.EyeFov.Viewport[eye] = ovrRecti(viewport);
+
+    return *this;
+}
+
+LayerEyeFov& LayerEyeFov::setRenderPoses(const Hmd& hmd) {
+    _layer.EyeFov.RenderPose[0] = hmd._ovrPoses[0];
+    _layer.EyeFov.RenderPose[1] = hmd._ovrPoses[1];
+
+    return *this;
+}
+
+LayerEyeFov& LayerEyeFov::setFov(const Hmd& hmd) {
+    _layer.EyeFov.Fov[0] = hmd._hmd->DefaultEyeFov[0];
+    _layer.EyeFov.Fov[1] = hmd._hmd->DefaultEyeFov[1];
+
+    return *this;
+}
+
+//----------------------------------------------------------------
+
+Compositor::Compositor(): _layers(), _wrappedLayers() {
+}
+
+typedef Corrade::Containers::EnumSet<LayerType> LayerTypes;
+
+Layer& Compositor::addLayer(const LayerType type) {
+    switch (type) {
+        case LayerType::Direct:
+            CORRADE_ASSERT(false, "LayerType::Direct is not implemented.", addLayerEyeFov());
+            break;
+        case LayerType::EyeFov:
+            return addLayerEyeFov();
+        case LayerType::EyeFovDepth:
+            CORRADE_ASSERT(false, "LayerType::EyeFovDepth is not implemented.", addLayerEyeFov());
+            break;
+        case LayerType::QuadHeadLocked:
+            CORRADE_ASSERT(false, "LayerType::QuadHeadLocked is not implemented.", addLayerEyeFov());
+            break;
+        case LayerType::QuadInWorld:
+            CORRADE_ASSERT(false, "LayerType::QuadInWorld is not implemented.", addLayerEyeFov());
+            break;
+    }
+    CORRADE_ASSERT_UNREACHABLE();
+}
+
+LayerEyeFov& Compositor::addLayerEyeFov() {
+    _wrappedLayers.push_back(std::move(std::unique_ptr<Layer>(new LayerEyeFov())));
+    _layers.emplace_back(&_wrappedLayers.back().get()->layerHeader());
+
+    return *static_cast<LayerEyeFov*>(_wrappedLayers.back().get());
+}
+
+Compositor& Compositor::submitFrame(const Hmd& hmd) {
+    ovrHmd_SubmitFrame(hmd.getOvrHmd(), 0, &hmd.getOvrViewScale(), _layers.data(), _layers.size());
+
+    return *this;
+}
+
+}}
--- a/src/Magnum/LibOvrIntegration/Compositor.cpp
+++ b/src/Magnum/LibOvrIntegration/Compositor.cpp
@@ -31,6 +31,7 @@
 #include "Magnum/LibOVRIntegration/Hmd.h"
 
 #include <OVR_CAPI_GL.h>
+#include <OVR_CAPI_Util.h>
 
 namespace Magnum { namespace LibOvrIntegration {
 
@@ -49,11 +50,27 @@ Layer& Layer::setEnabled(bool enabled) {
 }
 
 //----------------------------------------------------------------
+LayerDirect::LayerDirect(): Layer(LayerType::Direct) {
+}
+
+LayerDirect& LayerDirect::setColorTexture(const int eye, const SwapTextureSet& textureSet) {
+    _layer.Direct.ColorTexture[eye] = &textureSet.getOvrSwapTextureSet();
+
+    return *this;
+}
+
+LayerDirect& LayerDirect::setViewport(const int eye, const Range2Di& viewport) {
+    _layer.Direct.Viewport[eye] = ovrRecti(viewport);
+
+    return *this;
+}
+
+//----------------------------------------------------------------
 
 LayerEyeFov::LayerEyeFov(): Layer(LayerType::EyeFov) {
 }
 
-LayerEyeFov& LayerEyeFov::setColorTexture(const int eye, SwapTextureSet& textureSet) {
+LayerEyeFov& LayerEyeFov::setColorTexture(const int eye, const SwapTextureSet& textureSet) {
     _layer.EyeFov.ColorTexture[eye] = &textureSet.getOvrSwapTextureSet();
 
     return *this;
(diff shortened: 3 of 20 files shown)


mosra:
Thanks a lot for this effort. Great stuff, as I said already before.

Just a few things before you proceed further so you don't have to fix too much stuff later:

- To match naming policy of the rest of the project, I'd like to have `LibOvrIntegration` instead of `LibOVRIntegration` (you already did use `Hmd` instead of `HMD` which is great). But keep the `FindOVR.cmake` as is so it matches name of the DLL. Changing filename and directory case is rather tricky with Git on Windows, I'm able to help with that if you need.
- There's no need to have `void` in parameter-less function signatures (like `Layer::layerType(void)`). IIRC, this is some leftover from C, where the distinction actually had some meaning (but it doesn't in C++ and shorter is better).
- The `inline` keyword is implicit and thus redundant for [functions defined inside class body](http://en.cppreference.com/w/cpp/language/inline) -- little known feature and I don't blame you for not knowing it (but again, shorter is better).
- Please use either (const) references or `std::unique_ptr` instead of raw pointers and manual deletion, particularly when returning heap-allocated instances from functions to avoid accidental memory/resource leaks. The array of `Texture2D` pointers in `SwapTextureSet` is a tricky one, I have something in mind, but that needs to have some library support from my side.

That's for now, hopefully I didn't overwhelm you with the requests (sorry) :)

Squareys:
> To match naming policy [...]

I agree, I will just do a soft reset instead of a rebase then an recommit everything with the renamed directories, shouldn't be a problem.

> The inline keyword is implicit and thus redundant for functions defined inside class body

Good to know, thank you.

> Please use either (const) references or std::unique_ptr instead of raw pointers

In the cases where order of deletion matters, is it "clean" to use `unique_ptr.reset()` to delete?

> That's for now, hopefully I didn't overwhelm you with the requests (sorry) :)

No problem, I learn alot from this kind of feedback :)

mosra:
> I agree, I will just do a soft reset instead of a rebase then an recommit everything with the renamed directories, shouldn't be a problem.

The workaround that did the job for me was to rename in two steps (two commits that would be squashed later), first renaming the files/directories to completely different names and then back with proper case. Manual recommiting is not needed.

> In the cases where order of deletion matters, is it "clean" to use unique_ptr.reset() to delete?

Can you point me to the code where this is the case? Couldn't reordering the members help? (they are destructed in reverse order of declaration, last declared one is destructed first)

Squareys:
> Can you point me to the code where this is the case? 

https://github.com/Squareys/magnum-examples/blob/ovr-example/src/ovr/OVRExample.cpp#L190-L193 Was the code snipped I had in mind, can be solved through reordering, though, as you pointed out.

mosra:
Added the infrastructure for converting quaternions (mosra/magnum@94c4225165f72b970a67311cbcc40720c6595a27), dual quaternions (mosra/magnum@52434fb55c81df78ea3c57eb3a829cbc8cdc7a9d) and ranges (mosra/magnum@562d71dbe677c3dbd6427982895a62ce67dc9758) so you can finish the conversion utilities. The usage should be similar to what you have already done (see the test cases for hints). Funnily enough the new code revealed a GCC bug, so it took me much longer than I would like.

If I understand correctly, the `ovrPosef` class is just an orientation + position, so using dual quaternion instead of a matrix might be better idea in terms of memory efficiency and performance (and also way simpler conversion).

Squareys:
> If I understand correctly, the ovrPosef class is just an orientation + position, so using dual quaternion instead of a matrix might be better idea [...]

Yes, that's right. I never used dual quaternions before, but I'll figure it out.

> Added the infrastructure for [...]

Thanks! That'll make things alot cleaner.

I need to concentrate on my studies for a bit, so sorry in advance that there will be a little pause in progress, I hope to get back to this next weekend, though, maybe I'll even find some time in between aswell.

mosra:
Sure, you did amazing amount of work already :)

You don't need to understand the internal structure of dual quaternions to use them, they can be multiplied just like matrices and created from quaternion/translation pair like this:

```cpp
Quaternion rotation;
Vector3 translation
DualQuaternion a = DualQuaternion::translation(translation)*DualQuaternion{rotation};
```

Conversion to matrices is then done via the obvious [toMatrix()](http://mosra.cz/blog/magnum-doc/classMagnum_1_1Math_1_1DualQuaternion.html#a96659593dca1315f6b4b93be7d8e1d75). Also, for the example, scene graph can (but doesn't need to be) be templated on dual quaternion type instead of matrices, just use [SceneGraph::DualQuaternionTransformation](http://mosra.cz/blog/magnum-doc/classMagnum_1_1SceneGraph_1_1BasicDualQuaternionTransformation.html) instead of `SceneGraph::MatrixTransformation3D`.

Squareys:
Alright, I think I the code should now pretty much respect your requests. The test for converting the DualQuaternions still fails, though. (The `Texture2D** _textures;` is still there, also.)

What is your opinion on the includes? I had had problems with QtCreator opening the installed headers instead of the local ones, so includes of LibOvrIntegration headers are currently all relative.

Also, I should probably add some more tests for the other code aswell. I saw there is a AbstractGLTester or similiar in Magnum. That might come in handy.

modules/FindMagnumIntegration.cmake
```diff
@@ -116,7 +116,16 @@ foreach(component ${MagnumIntegration_FIND_COMPONENTS})
...
+    elseif(${component} STREQUAL LibOVR)
+		find_package(OVR)
+		if(OVR_FOUND)
+			set(_MAGNUM_${_COMPONENT}INTEGRATION_LIBRARIES ${OVR_LIBRARIES})
+            set(_MAGNUM_${_COMPONENT}INTEGRATION_INCLUDE_PATH_NAMES HmdPose.h)
+            set(_MAGNUM_${_COMPONENT}INTEGRATION_INCLUDE_DIRS ${PVR_INCLUDE_DIRS})
+		else()
+			unset(MAGNUM_${_COMPONENT}INTEGRATION_LIBRARY)
+		endif()
+	endif()
```
mosra: Looks like you mixed up indentation with tabs and spaces by accident.

modules/FindOVR.cmake
```diff
@@ -0,0 +1,25 @@
...
+endif()
+
+set(OVR_LIBRARIES
+	${OVR_LIBRARY}
+)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(OVR OVR_LIBRARY OVR_LIBRARIES OVR_INCLUDE_DIR)
+
+message(STATUS "OVR_FOUND = " ${OVR_FOUND})
```
mosra: These two status messages shouldn't be needed (`find_package_handle_standard_args()` prints a message on success/failure anyway).

src/Magnum/LibOVRIntegration/Conversion.h
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * @brief Wrap an ovrTexture as @ref Texture2D.
+ * @param texture @ref ovrTexture to wrap.
+ * @return texture as @ref Texture2D.
+ */
+Texture2D wrap(const ovrTexture& texture);
+
+}} /* namespace Magnum::LibOvrIntegration */
+
+#endif /* Magnum_LibOvrIntegration_Conversion_h */
```
mosra: Yes, I know that QtCreator adds these comments at namespace ending braces and `#endif` automatically, but I'd prefer to not have them, as it adds another burden when doing refactoring/file renaming later.

src/Magnum/LibOvrIntegration/Compositor.h
```diff
@@ -0,0 +1,217 @@
...
+        return _TYPE;
+    }
+
+    /** @brief The underlying ovrLayerHeader. */
+    const ovrLayerHeader& layerHeader() const {
+        return _layer.Header;
+    }
+
+protected:
+    ovrLayer_Union _layer;
```
mosra: (This is also rather QtCreator's fault -- I was unable to persuade it to do otherwise myself.)

I'm indenting the `public:`/`private:` access specifiers with 4 spaces and the contents inside with another 4 spaces (it might be wasteful, but I did that for whole project, so just for consistency).

src/Magnum/LibOvrIntegration/Context.h
```diff
@@ -0,0 +1,145 @@
...
+*/
+
+/** @file
+ * @brief Class @ref Magnum::LibOvrIntegration::OVRTextureUtil,
+ *        enum @ref Magnum::LibOvrIntegration::LibOvrContextStatusFlag
+ *
+ * @author Jonathan Hale (Squareys)
+ */
+
+#include <bits/unique_ptr.h>
```
mosra: `#include <memory>` ;)

src/Magnum/LibOvrIntegration/Hmd.h
```diff
@@ -0,0 +1,233 @@
...
+     * @brief Increment to use the next texture in the set.
+     * @return Reference to self (for method chaining)
+     */
+    SwapTextureSet& increment() {
+        _swapTextureSet->CurrentIndex = (_swapTextureSet->CurrentIndex + 1) % _swapTextureSet->TextureCount;
+        return *this;
+    }
+
+    /** @brief The underlying ovrSwapTextureSet. */
+    ovrSwapTextureSet* getOvrSwapTextureSet() const {
```
mosra: I would return reference here (assumption is that the member is never `nullptr` anyway).

src/Magnum/LibOvrIntegration/HmdEnum.h
```diff
@@ -0,0 +1,105 @@
...
+    HasMirrorTexture = 1 << 0,
+    /** The hmd was created as a debug hmd (without real hardware). */
+    Debug = 1 << 1
+};
+
+/**
+@brief Hmd tracking capabilities flags
+*/
+typedef Containers::EnumSet<HmdStatusFlag> HmdStatusFlags;
+
```
mosra: There should be also the following lines:

```cpp
CORRADE_ENUMSET_OPERATORS(HmdCapabilities)
CORRADE_ENUMSET_OPERATORS(HmdTrackingCapabilities)
CORRADE_ENUMSET_OPERATORS(HmdStatusFlags)
```

Otherwise code like `HmdCapability::LowPersistence|HmdCapability::NoVSync` wouldn't compile.

src/Magnum/LibOvrIntegration/LibOvrIntegration.h
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+/** @file
+ * @brief Forward declarations for the Magnum::LibOvrIntegration namespace.
+ *
+ * @author Jonathan Hale (Squareys)
+ */
+
+#include <Magnum/Magnum.h>
+#include "Conversion.h"
```
mosra: The purpose of forward declaration headers is to be as small as possible to have no negative impact on compilation times (just forward declarations and nothing else), so I'd remove the `#include "Conversion.h"` line.

src/Magnum/LibOvrIntegration/visibility.h
```diff
@@ -0,0 +1,39 @@
...
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+#include <Corrade/Utility/VisibilityMacros.h>
+
+#ifdef MagnumLibOvrIntegration_EXPORTS
+    #define MAGNUM_LibOvrIntegration_EXPORT CORRADE_VISIBILITY_EXPORT
```
mosra: Sorry, I'm now really nitpicking, but rather `MAGNUM_LIBOVRINTEGRATION_EXPORT` here (macros (except header guards) are always uppercase to be distinguishable from other stuff).

src/Magnum/LibOVRIntegration/Conversion.h
```diff
@@ -0,0 +1,157 @@
...
+#include <Magnum/Texture.h>
+#include <Magnum/Math/Range.h>
+#include <Magnum/Math/Vector.h>
+#include <Magnum/Math/Matrix4.h>
+#include <Magnum/Math/Quaternion.h>
+#include <Magnum/Math/DualQuaternion.h>
+
+namespace Magnum { namespace Math { namespace Implementation {
+
+// ovrSizei
```
mosra: Please use `/* .. */` comments rather than `// ...` (except in Doxygen blocks, because these comments cannot be nested). 

(Explanation: I have such policy that I'm using `// ...` *exclusively* for quick-and-dirty coding (commenting-out some part of code, short-lived todo/fixme comments etc.) so when I see it in a commit, it subconsciously triggers a warning that there is something that shouldn't be here.)

src/Magnum/LibOVRIntegration/Test/IntegrationTest.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+// define missing == operators for LibOVR types
+bool operator== (const ovrSizei& a, const ovrSizei& b);
+bool operator== (const ovrVector2i& a, const ovrVector2i& b);
+bool operator== (const ovrVector2f& a, const ovrVector2f& b);
+bool operator== (const ovrVector3f& a, const ovrVector3f& b);
+bool operator== (const ovrRecti& a, const ovrRecti& b);
+bool operator== (const ovrMatrix4f& a, const ovrMatrix4f& b);
+bool operator== (const ovrQuatf& a, const ovrQuatf& b);
+bool operator== (const ovrPosef& a, const ovrPosef& b);
```
mosra: If you wrap the operators in an anonymous namespace, these forward declarations won't be needed (If I'm guessing correctly, you added them to avoid compiler warning, right?).

modules/FindMagnumIntegration.cmake
Squareys: Indeed, code formatting has not been done yet, though.

src/Magnum/LibOVRIntegration/Test/IntegrationTest.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    return true;
+}
+
+bool operator== (const ovrQuatf& a, const ovrQuatf& b) {
+    if (a.x != b.x) return false;
+    if (a.y != b.y) return false;
+    if (a.z != b.z) return false;
+    if (a.w != b.w) return false;
+    return true;
```
mosra: Well, thinking about these, I think it might be better to drop the operators and do the comparisons directly in the test cases, as it would do proper fuzzy comparison for floats, will verbosely print what's wrong when the test fails and won't break if next release of Oculus SDK adds the comparison operators itself. So in the test case below:

```cpp
ovrQuatf c{a};
CORRADE_COMPARE(c.x, b.x);
CORRADE_COMPARE(c.y, b.y);
CORRADE_COMPARE(c.z, b.z);
CORRADE_COMPARE(c.w, b.w);
```

You can call the `CORRADE_COMPARE` macro in a loop, so that would work fine also in the matrix case.

src/Magnum/LibOVRIntegration/Conversion.h
Squareys: Actually, I added those :D And, yes, I also had to change it a few times already. In that case I will remove them again.

src/Magnum/LibOvrIntegration/Compositor.h
Squareys: Indeed, Qt did that, but again, I did not do code formatting yet. (See Todos above)

src/Magnum/LibOvrIntegration/Context.h
Squareys: I hoped to get rid of some big includes that way. I guess that is not recommended?

src/Magnum/LibOvrIntegration/HmdEnum.h
Squareys: Ah, good to know, I wondered why that didn't work already.

src/Magnum/LibOvrIntegration/LibOvrIntegration.h
Squareys: Will do.

src/Magnum/LibOvrIntegration/visibility.h
Squareys: Alright, I request change of https://github.com/mosra/magnum-integration/blob/master/src/Magnum/BulletIntegration/visibility.h for consistency then ;)

src/Magnum/LibOVRIntegration/Conversion.h
Squareys: Alright, I thing there are alot more of these, I will go through all the code again anyway, so changing the comments is not that much more work anyway.

src/Magnum/LibOvrIntegration/Hmd.h
Squareys: I think there was a reason why I left this one, but I'll check again.

src/Magnum/LibOvrIntegration/visibility.h
mosra: Um... (yeah, sorry, it's complicated), the `MagnumLibOvrIntegration_EXPORTS`/`MagnumBulletIntegration_EXPORTS` macro is provided by CMake, its name is hardcoded to `<library-filename>_EXPORTS` and I can't do much about that. 

I meant the `MAGNUM_LibOvrIntegration_EXPORT` one, which is then used in `Compositor.h` etc.

src/Magnum/LibOvrIntegration/Context.h
mosra: Yeah, I know what you mean, but sadly this won't work in libc++ (OSX) and with MSVC (both have `unique_ptr` defined right in `<memory>` header).

src/Magnum/LibOvrIntegration/visibility.h
Squareys: Ah, I see, sorry. What's with include guards? They should match case of folder an filenames, right?

mosra: Yup :)

mosra:
Okay, there won't be more annoying coding style comments from me, I promise :)

I'll look at the includes, it works for without a problem in other projects, so maybe there's something wrong just in this repo. How's the test failing? (Maybe I was wrong about how the dual quaternion should be initialized).

The `AbstractOpenGLTester` currently doesn't work on Windows (winapi requires me to pass some data through the constructor and I need to work around that somehow), will try to fix that soon.

I have one last note about the `LibOvrContext` class. It apparently needs to be some sort of a singleton, but I think it could be done without the `initialize()`/`finalize()` methods. Looking at the example, it's now like this:
```cpp`
// LibOvrContext is created on heap statically before entering main()
{
    LibOvrContext::get()->initialize();

    // code that accesses the context through LibOvrContext::get() ... 

    LibOvrContext::get()->finalize();
}
```

I propose changing that to the following:
```cpp
{
    LibOvrContext _context; // as a private member of the application class

    // (unchanged) code that accesses the context through LibOvrContext::get() ... 

    // _context gets automatically destroyed when application is destructed
}
```

The static `_instance` variable will be initialized to `nullptr`, the constructor then will point it to itself (and assert that there was no other instance before), destructor sets it back to `nullptr` (and assert that it was equal to `this`) and `get()` method would return a reference instead of pointer (and assert that static instance is not `nullptr`). Should be a bit easier for the end user.

mosra:
Just checked the absolute includes and they appear to work in QtCreator on Windows (e.g. ctrl-clicking on `#include "Magnum/LibOvrIntegration/Conversion.h"` inside `Conversion.cpp` file was working), if you install them to the same location as the rest of Magnum then they should work also when installed.

Huh... it looks like the first two commits still have the `LibOVRIntegration` directory name, the case insensivity is painful :)

Squareys:
> Okay, sorry again about annoying coding style comments :)

I think I got used to it already ;)

> How's the test failing?

`.translation()` on the instance does not give me the values I set in the first place with `DualQuaternion::translation`.

> I have one last note about the LibOvrContext class. [...]

I had remembered that `ovr_Initialize()` needed to be called before any OpenGL call. That was removed in the current sdk version, though. And in this case your suggestion is alot cleaner, of course.

In your suggestion though, even accessing `_context` directly from Application would work, right?

> Huh... it looks like the first two commits still have the LibOVRIntegration directory name, case insensitive filesystems are painful :)

Oh, well. I am on Linux currently, I might just cleanup the directories here later then.

> Just checked the absolute includes [...]

Ctrl-clicking did work, yes, but it kept sending me to the installed files, causing a bit of confusion when I edited them and the local files did not change. I will try changing back the includes to be absolute and see if that problem persists.

Squareys:
Folder renaming complete, I guess the squashing should happen under linux aswell.

mosra:
> `.translation()` on the instance does not give me the values I set in the first place with `DualQuaternion::translation`.

I *think* this is related to the fact that the rotation quaternion you are using in the test case is not normalized. Try this instead:
```cpp
DualQuaternion{Quaternion{{0.1, 0.2, 0.3}, 1.0}.normalized()}
```

I need to look into the math again, apparently I completely forgot everything about dual quaternions :) I should also put some assertion somewhere to catch this case.

> I had remembered that `ovr_Initialize()` needed to be called before any OpenGL call.

Even before creating GL context? Huh... Also, doesn't the SDK mess somehow with GL state so it needs to be accounted for on our side using `Context::resetState()`? Or are they reverting back all the state that they touch (slow...)?

> In your suggestion though, even accessing `_context` directly from Application would work, right?

Yup. Best case of course would be if there is no need for the global instance.

> it kept sending me to the installed files

Well, yeah :/ I have this exact problem with KDevelop, somehow system-wide includes are more important than the local ones. I'll try to investigate, maybe it is somehow related to order of include directories in CMake.

Squareys:
> I think this is related to the fact that the rotation quaternion you are using in the test case is not normalized.

Yeah, pretty sure that's it. As far as I know, non-normalized quaternions cannot be used for rotations.

> Even before creating GL context?

You might have missed the next sentence in which I explained that this not the case in the current SDK version anymore and won't be so in the future.

> Yup. Best case of course would be if there is no need for the global instance.

Well, it doesn't really make sense to have more than one context, also that would probably mess up the `ovr_Initialize()` and `ovr_Shutdown()` calls. So, I would keep it as a singleton.

src/Magnum/LibOvrIntegration/Hmd.h
Squareys: There wasn't, I just need to keep the `_swapTextureSet` as a pointer, because `ovrHmd_CreateSwapTextureSetGL()` expects `ovrSwapTextureSet**`.

Squareys:
Some comments got lost, since they where attatched to diffs:

@mosra wrote:
> Please use /* .. */ comments rather than // ... (except in Doxygen blocks, because these comments cannot be nested).

> (Explanation: I have such policy that I'm using // ... exclusively for quick-and-dirty coding (commenting-out some part of code, short-lived todo/fixme comments etc.) so when I see it in a commit, it subconsciously triggers a warning that there is something that shouldn't be here.)

I think, this probably needs an update: http://mosra.cz/blog/corrade-doc/corrade-coding-style.html#corrade-coding-style-documentation

@mosra wrote:
> Well, thinking about these, [the ovr structure comparison operators] I think it might be better to drop the operators and do the comparisons directly in the test cases, as it would do proper fuzzy comparison for floats, will verbosely print what's wrong when the test fails and won't break if next release of Oculus SDK adds the comparison operators itself. So in the test case below:

> ```
ovrQuatf c{a};
CORRADE_COMPARE(c.x, b.x);
CORRADE_COMPARE(c.y, b.y);
CORRADE_COMPARE(c.z, b.z);
CORRADE_COMPARE(c.w, b.w);
```

> You can call the CORRADE_COMPARE macro in a loop, so that would work fine also in the matrix case.

Alright.

**Comment by mosra** [comment]
> I think, this probably needs an update: http://mosra.cz/blog/corrade-doc/corrade-coding-style.html#corrade-coding-style-documentation

I know, yeah, it's confusing *a lot*, really, and it's bugging me all the time, but I was unable to persuade Doxygen to accept anything else -- writing `/** /* */ */` completely breaks syntax highlighting in _any_ editor (even though Doxygen somehow accepts that). That would be somehow bearable, though (abandon all editor sanity and autocompletion benefits), but writing `/** */` inside Doxygen `@code` block is even more impossible because Doxygen then tries to parse it as an actual documentation block and that part then magically gets lost from the docs completely :D

I'll try to do something about it, though.

**Comment by Squareys** [comment]
> I'll try to do something about it, though.

You *could* use `/// /*  */` as an exception for documenting documentation.

**Comment by Squareys** [comment]
So, should
`int s; //< short line.`
be
`int s; /*< short line. */`
?

**Comment by mosra** [comment]
Yup. I'm on it :)

**Comment by Squareys** [comment]
I'm on the finish stretch here.

I will probably finish the magnum-ovr example before this can be merged. I will take the working examples for the documentation from there, when that is done.

**Comment by Squareys** [comment]
Last TODOs left:  Finish doc++ and then squash commits. Will finish https://github.com/mosra/magnum-examples/pull/10 first, though.

File: src/Magnum/LibOvrIntegration/Compositor.cpp [hunk 13]
```diff
@@ -0,0 +1,125 @@
...
+    }
+    CORRADE_ASSERT_UNREACHABLE();
+}
+
+LayerEyeFov& Compositor::addLayerEyeFov() {
+    std::unique_ptr<Layer> wrapped(new LayerEyeFov());
+    _wrappedLayers.push_back(std::move(wrapped));
+    _layers.push_back(&wrapped->layerHeader());
+
+    return *static_cast<LayerEyeFov*>(_wrappedLayers.back().get());
```
Squareys: Well, I kinda screwed up this part. I will add some tests for construction and deconstruction of some of the classes, I don't really trust my code anymore.

mosra: Just swapping the two `push_back()` calls should do it, no? :) As far as I can tell, tests for this would require active GL context. I'll push the needed changes for `AbstractOpenGLTester` this evening, hopefully.

Oh, looking above, the `CORRADE_ASSERT(true, ...)` calls should probably be `false`, otherwise they won't do anything. I'd remove the `default:` case as well, because then some compilers would be able to issue a warning when there's some unhandled enum value (very valuable warning, in my opinion). ... um, and, the `_TYPE` member should be rather `_type`, underscore followed by uppercase letter is reserved by the c++ standard.

Sorry, got carried away again :)

Squareys: I already swapped those calls, yes (haven't pushed that yet, though).

mosra:
Added support for GL tests on Windows in mosra/magnum@88f04152905fb4dc3aefe2ed7fa81681e24f15e1, in case you would want to do them. And now off for the weekend :)

Squareys:
Big squash incoming! I will add the doc in separate commits instead of fixups, if you are okay with that.

mosra:
Sure, adding documentation to existing commits is no fun :)

Code looks great, I never imagined it would get so large. One last thing, could you rename the `Conversion{.h,cpp}` back to `Integration.{h,cpp}`? I know the name is not exactly clear, but let's stay consistent with Bullet integration lib.

Squareys:
@mosra Alright, will do.

I'm currently wrapping another compositor layer for debugging (LayerDirect), and it turns out, that all of the methods needed are already implemented in LayerEyeFov.
I could let LayerEyeFov inherit from LayerDirect, but then the method chaining would get messed up (calls to `setViewport()` for example would return a reference to LayerDirect instead of LayerEyeFov when called from an instance of the later.) 
What would you recommend? 

- Is it okay to have a LayerEyeFov method return a reference to LayerDirect "for method chaining", or
- is that so bad that you would recommend code duplication (or using macros to define those methods 2-5 times in different subclasses of Layer)
- Or templates? Could it work defining the six method templates and then creating the required instances of those templates in the Layer subclasses?
- Or maybe just remove method chaining for Layer?

The inheritance doesn't really serve any more purpose than not having to rewrite the methods.

Squareys:
While I was at it, I implemented all the layer types after all. I went with the code duplication solution, since inheritance is not actually desirable with the layers and there weren't that many shared methods overall.

And back to the example.

Squareys:
@mosra I declare this ready for review aswell :) 

You may want to cherry-pick the bullet doc fix (I could also create a new pull request for that, but it seemed a little overkill for one line change), to keep things clean. I can then rebase this onto master make it automatically mergable again.

mosra:
No need to be so strict about pull requests, I'll merge (or fast-forward) everything in one go.

Squareys:
@mosra Okay, ready for review/merge once again.

PR #10, opened 2015-11-15
Update to Oculus SDK 0.8.0.0-beta

Squareys:
Hi @mosra !

This is an early PR to let you know I'm working on this. I actually found some old changes I forgot about in my local repository `:D`

TODOs:
 - [x] `ovr_Detect` function is back
 - [x] ovr_GetSessionStatus
 - [x] check error enum for hidden changes (yes, there were some)
 - [x] update cmake modules in examples
 - [x] investigate latencyMarker flag in ovrTrackingState
 - [x] ovr_ConfigureTracking is not required, but is there a use case where one would not want full tracking capabilities (just in case there is this use case, the method will be left, doc/example will be adapted)
 - [x] ovrLayerHudMode
 - Support for Oculus Touch will be prosponed until I can make use of it.

*(those TODOs were written late at night, my appologies for anything weird that may have resulted from that.)*

Greetings,
Squareys


mosra:
Oh. Wow. Slow down, I can't keep up! :)

Now I have genuine fear that we get banned from the CIs soon :D

Squareys:
I will try to batch more commits together before pushing, to grant the CIs a break every now and then.

mosra:
I'm not complaining, it's just that the AppVeyor builds seemed to take three times longer than before (12 minutes vs. 4 minutes), like if we were throttled because of excessive activity :)

src/Magnum/OvrIntegration/Compositor.h
```diff
@@ -27,7 +27,11 @@
 */
 
 /** @file
- * @brief Class @ref Magnum::OvrIntegration::Compositor, @ref Magnum::OvrIntegration::Layer, @ref Magnum::OvrIntegration::LayerDirect, @ref Magnum::OvrIntegration::LayerEyeFov, @ref Magnum::OvrIntegration::LayerEyeFovDepth, @ref Magnum::OvrIntegration::LayerQuad, @ref Magnum::OvrIntegration::TimewarpProjectionDescription, enum @ref Magnum::OvrIntegration::LayerType
+ * @brief Class @ref Magnum::OvrIntegration::Compositor, @ref Magnum::OvrIntegration::Layer,
+ *    @ref Magnum::OvrIntegration::HeadLockableLayer, @ref Magnum::OvrIntegration::LayerDirect,
+ *    @ref Magnum::OvrIntegration::LayerEyeFov, @ref Magnum::OvrIntegration::LayerEyeFovDepth,
+ *    @ref Magnum::OvrIntegration::LayerQuad, @ref Magnum::OvrIntegration::TimewarpProjectionDescription,
+ *    enum @ref Magnum::OvrIntegration::LayerType
```
mosra: Sorry, but this will break Doxygen, we *have to* put `@brief`s on a single line (annoying, I know).

src/Magnum/OvrIntegration/Context.h
```diff
@@ -53,6 +53,17 @@ struct Error {
...
 /**
+ * @brief Result of @ref Context::detect(int).
+ */
+enum class OvrDetectResultFlag: Byte {
+    ServiceRunning = 1,
+    HmdConnected = 2
+};
+
+typedef Corrade::Containers::EnumSet<OvrDetectResultFlag> OvrDetectResult;
+CORRADE_ENUMSET_OPERATORS(OvrDetectResult)
```
mosra: Maybe naming this `OvrDetectResultFlags`? ... also, detect or detected?

src/Magnum/OvrIntegration/Hmd.cpp
```diff
@@ -160,16 +160,39 @@ bool Hmd::isDebugHmd() const {
...
 
 void Hmd::setDebugHudStereoMode(const DebugHudStereoMode mode) const {
-    ovr_SetInt(_hmd, OVR_DEBUG_HUD_STEREO_MODE, Int(mode));
+    ovr_SetInt(_session, OVR_DEBUG_HUD_STEREO_MODE, Int(mode));
+}
+
+Hmd::EyeCup Hmd::eyeCup() const {
+    char value[16];
+    ovr_GetString(_session, OVR_KEY_EYE_CUP, "A");
+    if(strcmp(value, "A")) {
```
mosra: `std::strcmp`, please. I guess there is a reason why they are using string and not enum? For example, wouldn't this be vendor-dependent or something?

src/Magnum/OvrIntegration/Hmd.h
```diff
@@ -191,6 +192,19 @@ Framebuffer::blit(mirrorFramebuffer,
...
+         * @see @ref Hmd::eyeCup()
+         */
+        enum class EyeCup: UnsignedShort {
+            A,
+            B,
+            C,
+            Unknown = 0xFF
+        };
+
+        /** Destructor. */
```
mosra:
```cpp
/**
 * @brief Eye cup
 * @see @ref eyeCup()
 */
```

The destructor doc is not needed as it doesn't bring any new information.

src/Magnum/OvrIntegration/Compositor.h
Squareys: Alright, we probably had discussed that back then when the code was first written. I will change it back.

 - [ ]

src/Magnum/OvrIntegration/Context.h
Squareys: Well it's the result of the `ovr_Detect` method, which makes this the result of Detect. See `static OvrDetectResult Context::detect(int)`

src/Magnum/OvrIntegration/Hmd.cpp
Squareys: I don't think the Oculus SDK will be used by anybody else than Oculus, I can return string, of course, but it's always A, B or C. Who knows what will be in the future, though.

- [ ]

src/Magnum/OvrIntegration/Hmd.h
Squareys: - [ ]

Squareys: The enum was removed in favor of return string in `Hmd::eyeCup()`

src/Magnum/OvrIntegration/Context.h
Squareys: In the SDK there is a struct of 2 bools which is called `ovrDetectResult`

Squareys: @mosra Would you be okay with keeping the naming as is? Would you suggest just returning the same structure/creating an own equivalent structure instead of using enums?

mosra: Sorry about a late reply, got busy once again ... 

Enum is great, *way* better than a structure of two bools (heh), but I feel like there is an inconsistency in the naming (the enum has `*Flag`, but the enum set doesn't). Elsewhere I was always having either `SomethingFlag` and `SomethingFlags` or tried very hard to invent something else so that is always easy to guess name of the other type. 

Hmm, what about `OvrDetectResult` for the enum and `OvrDetectResults` for the enumset?

Squareys: > Hmm, what about OvrDetectResult for the enum and OvrDetectResults for the enumset? (Yes, I know it's not consistent with ovrDetectResult.

No, that's great, very good compromise.

Squareys:
@mosra Once again, done from my perspective. Can potentially be merged with https://github.com/mosra/magnum-examples/pull/19 :)

src/Magnum/OvrIntegration/CMakeLists.txt
```diff
@@ -46,6 +46,10 @@ set(MagnumOvrIntegration_HEADERS
 
     visibility.h)
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CORRADE_CXX_FLAGS}")
+
+include_directories(${MAGNUM_INCLUDE_DIRS} ${MAGNUM_APPLICATION_INCLUDE_DIRS} ${MAGNUM_AUDIO_INCLUDE_DIRS} ${CMAKE_CURRENT_BINARY_DIR})
+
```
mosra: Why is this change here? These operations are already done in `src/CMakeLists.txt`.

src/Magnum/OvrIntegration/Hmd.cpp
```diff
@@ -26,6 +26,8 @@
 
 #include "Magnum/OvrIntegration/Hmd.h"
 
+#include <cstring>
+
```
mosra: I guess this is some leftover, not needed anywhere.

src/Magnum/OvrIntegration/Hmd.h
```diff
@@ -358,24 +370,110 @@ class MAGNUM_OVRINTEGRATION_EXPORT Hmd {
...
+        }
+
+        /** @brief Eye height set in the currently active Oculus profile */
+        Float eyeHeight() const {
+            return ovr_GetFloat(_session, OVR_KEY_EYE_HEIGHT, OVR_DEFAULT_EYE_HEIGHT);
+        }
+
+        /**
+         * @brief Inter populary distance set in the currently active Oculus
+         * profile
```
mosra: As noted previously, we have to put `@brief`s onto single line to make Doxygen parse them properly. Also, "interpupillary", I think :)

src/Magnum/OvrIntegration/Compositor.h
```diff
@@ -149,6 +149,35 @@ class MAGNUM_OVRINTEGRATION_EXPORT Layer {
         friend class Compositor;
 };
 
+class MAGNUM_OVRINTEGRATION_EXPORT HeadLockableLayer: public Layer {
```
mosra: Missing class documentation.

src/Magnum/OvrIntegration/Hmd.h
```diff
@@ -141,14 +138,11 @@ Framebuffer framebuffer{{}, textureSize};
...
 
 // setup depth attachment
-Image2D image(PixelFormat::DepthComponent, PixelType::UnsignedInt, textureSize, nullptr);
-
 Texture2D* depth = new Texture2D();
 depth->setMinificationFilter(Sampler::Filter::Linear)
       .setMagnificationFilter(Sampler::Filter::Linear)
       .setWrapping(Sampler::Wrapping::ClampToEdge)
       .setStorage(1, TextureFormat::DepthComponent24, textureSize)
-      .subImage(0, {{}, textureSize}, image);
```
mosra: The `;` should be on previous line then :)

src/Magnum/OvrIntegration/Test/EnumTest.cpp
```diff
@@ -115,6 +133,26 @@ void EnumTest::errorType() {
...
 }
 
+void EnumTest::ovrDetectResult() {
+    std::ostringstream out;
+    Debug(&out) << OvrDetectResult::HmdConnected;
+    CORRADE_COMPARE(out.str(), "OvrIntegration::OvrDetectResult::HmdConnected\n");
+
+    out.str("");
+    Debug(&out) << OvrDetectResult(-1);
+    CORRADE_COMPARE(out.str(), "OvrIntegration::OvrDetectResult::(invalid)\n");
```
mosra: :+1: awesome, you are testing really everything :)

src/Magnum/OvrIntegration/Hmd.h
```diff
@@ -358,24 +370,110 @@ class MAGNUM_OVRINTEGRATION_EXPORT Hmd {
...
         ovrPosef _ovrPoses[2];
         ovrVector3f _hmdToEyeViewOffset[2];
         ::ovrViewScaleDesc _viewScale;
 
-        ovrFrameTiming _frameTiming;
+        Double _predictedDisplayTime;
         ovrTrackingState _trackingState;
 
-        UnsignedInt _frameIndex;
+        long long _frameIndex;
```
mosra: Signed? Magnum has `Long`/`UnsignedLong` type which maps to `(unsigned) long long`. Shouldn't also the `incFrameIndex()` function have different return type then?

src/Magnum/OvrIntegration/Hmd.h
```diff
@@ -358,24 +370,110 @@ class MAGNUM_OVRINTEGRATION_EXPORT Hmd {
...
+         */
+        std::array<Float, 2> maxEyeToPlateDistance() const {
+            std::array<Float, 2> values{{0.0f, 0.0f}};
+            ovr_GetFloatArray(_session, OVR_KEY_MAX_EYE_TO_PLATE_DISTANCE, values.data(), 2);
+            return values;
+        }
+
+        /** @brief Eye cup set in the currently active Oculus profile */
+        std::string eyeCup() const {
+            return ovr_GetString(_session, OVR_KEY_EYE_CUP, "A");
```
mosra: I guess there is some really simple explanation, but why the `"A"` here? Is that some reasonable default?

src/Magnum/OvrIntegration/HmdEnum.h
```diff
@@ -184,6 +211,10 @@ enum class ErrorType: Int {
     MismatchedAdapters = ovrError_MismatchedAdapters,   /**< Chosen rendering adapters between client and service do not match */
     LeakingResources = ovrError_LeakingResources,       /**< Calling application has leaked resources */
     ClientVersion = ovrError_ClientVersion,             /**< Client version too old to connect to service */
+    OutOfDateOS = ovrError_OutOfDateOS,                 /**< The operating system is out of date. */
+    OutOfDateGfxDriver = ovrError_OutOfDateGfxDriver,   /**< The graphics driver is out of date. */
+    IncompatibleGPU = ovrError_IncompatibleGPU,         /**< The graphics hardware is not supported */
+    NoValidVRDisplaySystem = ovrError_NoValidVRDisplaySystem,   /**< No valid VR display system found. */
```
mosra: `IncompatibleOs`, `OutOfDateOs`, `IncompatibleGpu`, `NoValidVrDisplaySystem`, `UsbBandwidth`, `UsbEnumeratedSpeed` and `HmdFirmwareMismatch` so we have everything CamelCase to be consistent with the rest of Magnum. Looks like even Oculus can't get it consistent in a single enum :)

src/Magnum/OvrIntegration/CMakeLists.txt
Squareys: Ah, I missed those. I was really surprised how this missing made it to code review, should have investigated further, I guess. Will remove it again, sorry ;)

src/Magnum/OvrIntegration/Hmd.h
Squareys: oops :P Good catch.

src/Magnum/OvrIntegration/Test/EnumTest.cpp
Squareys: Yeah, that invalid thing is my most frequent source of copy-paste error, so back then I started doing that for all of the many OvrIntegration enums.

src/Magnum/OvrIntegration/Hmd.h
Squareys: Ah, I see, thanks for pointing that out. Yes, `incFrameIndex` should probably return the same type.

src/Magnum/OvrIntegration/Hmd.h
Squareys: I guess that is the default, and I probably took that from some defaults by the SDK itself, I'll check again, though, just to be on the safe side (in my oppinion "B" would be a more reasonable default, since it's the lense without sight correction)

Squareys:
@mosra Hey, you're back! :D I will take some time on the changes you proposed, probably wont find time before the weekend.

Squareys:
@mosra Alright, I finally got around to taking care of the flaws you pointed out! :)

mosra:
:+1: Merged.

I just rebased the commits to current master and did a single change -- removed redundant prefix from the `OvrDetectResult` enum. Hope that's okay.

Squareys:
> Hope that's okay.

No problem, of course, thanks!

PR #12, opened 2016-03-30
Update to Oculus SDK 1.3.0

Squareys:
Hello everybody!

The first non beta version of the Oculus SDK has finally been released and I'm working on updating OvrIntegration respectively.

TODO:
 - [x] Figure out what's going on with the layers (some removed, LayerEyeMatrix added)
 - [x] Adapt to mirror texture changes
 - [x] Make sure documentation is still correct
 - [ ] Other minor changes and renaming
   - [x] Changes in `ErrorType` enum
   - [x] Add `TrackingOrigin` enum and `Hmd::setTrackingOriginType()`
 - [x] Fix FindOVR.cmake (for MSVC)\*
 - [x] Wrap input related API
    - [x] Add `ControllerType` enum
    - [x] Get controller state
    - [x] Get tracker poses
 - [x] Rename `Hmd` to `Session`

\* Not tested yet. AppVeyor fails at downloading the Oculus SDK.

This should be done this Sunday.

Regards, 
Jonathan.


mosra:
Wow, awesome, you are continuously doing great job! :)

Didn't look into the changes yet, just an idea that came up on my mind -- could you try enabling the library on AppVeyor for MSVC 2015? It's explicitly set to `OFF` because there were linker problems before. I'm curious to see whether it works already or not.

Squareys:
Well, AppVeyor really does not seem to want to download the SDK...

Squareys:
In the last SDK update `ovrHmd` was renamed to `ovrSession`, which we could/should follow with `Hmd` (This is sane, since you require the session to get controller input state, which semantically does not make sense to get from the headset). Three options:
- Don't rename
- Rename, but keep an alias to not break outside code (who cares, I guess, others can do that themselves if  they must)
- Complete Rename of `Hmd` to `Session`

I'd prefer the last. What do you say, @mosra?

mosra:
Rename `Hmd` to `Session`.

We're not doing any more harm than Oculus itself when doing those renames, so that's okay.

Squareys:
@mosra I'm basically done here! I might stumble over the one or the other change in the SDK so there might be one or two more commits over the next week, but apart from that this should be complete.

mosra:
Great! :) Will go over it in the next few days and merge.

The AppVeyor build is failing with strange error, though. Never seen that before. Do you have an idea what could cause that?

Squareys:
@mosra Yes, it was because I passed an aligned structure by value instead if const ref. Should have fixed it :)

modules/FindOVR.cmake
```diff
@@ -39,28 +39,61 @@
 #   DEALINGS IN THE SOFTWARE.
 #
 
-set(LIBOVR_ROOT ${CMAKE_INSTALL_PREFIX})
-
 if(NOT OVR_SDK_ROOT)
-    message(WARNING "OVR_SDK_ROOT is not set. Will try to find headers and library in ${CMAKE_INSTALL_PREFIX}." )
-else()
-    set(LIBOVR_ROOT ${OVR_SDK_ROOT}/LibOVR)
+   find_path(OVR_SDK_ROOT OculusSDK)
```
mosra: Minor nitpick: accidental indentation with three spaces

modules/FindOVR.cmake
```diff
@@ -39,28 +39,61 @@
...
+endif()
+
+if(OVR_SDK_ROOT)
+    set(OVR_SDK_ROOT "${OVR_SDK_ROOT}/OculusSDK")
+
+    mark_as_advanced(OVR_SDK_ROOT)
```
mosra: Only variables with explicit `CACHE` setup or variables coming from `find_*()` commands need to be marked as advanced. All other variables don't go into the cache at all.

modules/FindOVR.cmake
```diff
@@ -39,28 +39,61 @@
...
+        elseif (MSVC_VERSION GREATER 1600)
+            set(MSVC_NAME "VS2012")
+        else()
+            set(MSVC_NAME "VS2010")
+        endif()
+
+        find_library(OVR_LIBRARY NAMES LibOVR HINTS "${LIBOVR_ROOT}/Lib/Windows/${MSVC_ARCH}/Release/${MSVC_NAME}")
+
+        unset(MSVC_ARCH)
+        unset(MSVC_NAME)
```
mosra: I would suggest using prefixes to avoid conflicts with user variables -- so e.g. `_OVR_MSVC_NAME` or something like that, leading underscore suggesting that this is an internal variable.

src/Magnum/OvrIntegration/HmdEnum.h
```diff
@@ -83,6 +86,183 @@ typedef Containers::EnumSet<HmdTrackingCapability> HmdTrackingCapabilities;
...
+    FloorLevel = ovrTrackingOrigin_FloorLevel,
+};
+
+/**
+@brief Tracker flag
+
+@see @ref TrackerFlags
+*/
+enum class TrackerFlag: Int {
+    /**< The sensor is present, else the sensor is absent or offline */
```
mosra: Superfluous `<` in the doc block (would cause the enum value to be undocumented).

src/Magnum/OvrIntegration/HmdEnum.h
```diff
@@ -83,6 +86,183 @@ typedef Containers::EnumSet<HmdTrackingCapability> HmdTrackingCapabilities;
...
+CORRADE_ENUMSET_OPERATORS(Buttons)
+
+/** @brief Bit mask of all buttons on the right Touch controller */
+constexpr Buttons R_BUTTON_MASK = Button::A | Button::B | Button::RThumb | Button::RShoulder;
+
+/** @brief Bit mask of all buttons on the left Touch controller */
+constexpr Buttons L_BUTTON_MASK = Button::X | Button::Y | Button::LThumb | Button::LShoulder;
+
+/** @brief Bit mask of buttons used by Oculus Home */
+constexpr Buttons PRIVATE_MASK = Button::VolUp | Button::VolDown | Button::Home;
```
mosra: I'm not sure about the naming convention and lack of scoping here.

Just an idea -- would it make sense to have the mask part of the Button enum?

src/Magnum/OvrIntegration/HmdEnum.h
```diff
@@ -224,30 +424,55 @@ enum class ErrorType: Int {
...
+    MetricsDuplicateApp = ovrError_MetricsDuplicateApp,
+    MetricsNoEvents = ovrError_MetricsNoEvents,
+    MetricsRuntime = ovrError_MetricsRuntime,
+    MetricsFile  = ovrError_MetricsFile,
+    MetricsNoClientInfo = ovrError_MetricsNoClientInfo,
+    MetricsNoAppMetaData = ovrError_MetricsNoAppMetaData,
+    MetricsNoApp = ovrError_MetricsNoApp,
+    MetricsOafFailure = ovrError_MetricsOafFailure,
+    MetricsSessionAlreadyActive = ovrError_MetricsSessionAlreadyActive,
+    MetricsSessionNotActive = ovrError_MetricsSessionNotActive,
```
mosra: Doc blocks missing here.

src/Magnum/OvrIntegration/Session.h
```diff
@@ -49,22 +49,173 @@
...
 /**
-@brief Swap texture set
+ * @brief A full pose (rigid body) configuration with first and second derivatives.
+ *
+ * Body refers to any object for which ovrPoseStatef is providing data.
+ * It can be the HMD, Touch controller, sensor or something else. The context
+ * depends on the usage of the struct.
+ */
+class MAGNUM_OVRINTEGRATION_EXPORT PoseState {
+public:
```
mosra: Please indent also the `public:`/`private:` labels to be consistent with the rest of the codebase.

src/Magnum/OvrIntegration/Session.h
```diff
@@ -49,22 +49,173 @@
...
+
+    static PoseState& wrap(::ovrPoseStatef& state) {
+        return reinterpret_cast<PoseState&>(state);
+    }
+    static const PoseState& wrap(const ::ovrPoseStatef& state) {
+        return reinterpret_cast<const PoseState&>(state);
+    }
```
mosra: Doc blocks missing.

mosra:
:+1: 

The only concern I have are the masks for `Buttons` and `Touches`, the rest are just minor issues. First-class work again! :)

modules/FindOVR.cmake
Squareys: In this special case, I would do explicit cache setup to enable the user to specify a custom `OCULUS_SDK_ROOT`. Would that make sense?

modules/FindOVR.cmake
Squareys: Will do, makes sense :+1:

src/Magnum/OvrIntegration/HmdEnum.h
Squareys: This is harder than expected, gcc will complain about "use of incomplete type" for `Buttons`. I'm trying to figure out how to do this, but will fix the other issues first.

src/Magnum/OvrIntegration/HmdEnum.h
Squareys: Ah, that's because there was no doc in the SDK :D

Squareys: I just found lwjgl also wraps the Oculus SDK and they document these as `Error codes (ovrErrorType)`...

Some of these I can't even imagine what they are supposed to mean.

src/Magnum/OvrIntegration/HmdEnum.h
Squareys: Also see [this stackoverflow discussion](http://stackoverflow.com/questions/11928089/static-constexpr-member-of-same-type-as-class-being-defined)

Squareys:
@mosra Alright, I fixed all the minor issues. The solution you proposed for Buttons and Touches does not quite work as hoped. I understand why the initial solution is not nice, so I will try to find another solution for that.

src/Magnum/OvrIntegration/HmdEnum.h
mosra: Huh :D What about just braindead repeat of the enum name until Oculus provides something better? E.g. `MetricsNoAppMetaData` -> `/**< No app metadata metrics */`

src/Magnum/OvrIntegration/HmdEnum.h
mosra: Oh crap.

Since the type isn't really needed (but still would be nice to have it there for Doxygen):
```cpp
struct Buttons: Containers::EnumSet<Button> {
    constexpr 
        #ifdef DOXYGEN_GENERATING_OUTPUT
        Buttons
        #else
        Containers::EnumSet<Button>
        #endif
        LMask = ...;

    using EnumSet::EnumSet; // inherit the constructors
};
```

modules/FindOVR.cmake
mosra: Then you need to so something like `set(OVR_SDK_ROOT "<some default path>" CACHE PATH "OculusVR SDK root")` to make it visible in the GUI and avoid overwriting it with a different value later to reduce confusion.

src/Magnum/OvrIntegration/HmdEnum.h
Squareys: Sure, will do that tomorrow :)

src/Magnum/OvrIntegration/HmdEnum.h
Squareys: Done :)

mosra:
Merged in ff7b924edbf445f24b49a022cb0d7a4d45ba4752, sorry it took so long. I did a rough cleanup and review of the code, but because the PR was already quite big, I put the comments into a separate issue: #13. If you could go over it once you have time, that would be great. It's not extreme high priority, though, you have other things to focus on :)

Issue #13, opened 2016-05-12
OVR SDK integration cleanup


mosra:
Hi @Squareys,

While merging #12 I went over the code and did a rough formatting and documentation cleanup. Few things were not clear to me as I'm not very familiar with the API, so once you get some spare time, can you go through the following issues?

*  The HMD -> Session rename is not fully done, some function parameters are still named `hmd` though the type is now `Session`
*  Similarly with texture set -> texture swap chain, types don't match function parameter names and docs
*  There are some types still named `Hmd` but used in the Session class, is that okay?
*  Is `OvrIntegration/HmdEnum.h` still named correctly? What about just `OvrIntegration/Enums.h`?
*  The documentation of Session class functions mentions things like

    > ```
    >          * If you are using `ovrTrackerPoses` then you will need to call
    >          * `ovr_GetTrackerPose` after this, because the sensor position(s) will
    >          * change as a result of this.
    > ```

    More occurences elsewhere. I think the docs should mention the wrapper APIs and not the original OVR SDK APIs.

*   The `HmdTrackingCapability` and `HmdTrackingCapabilities` don't seem to be used anywhere and reference removed `Hmd::configureTracking()` function. Similarly with `StatusFlag` and `StatusFlags`. Should they be removed also?
*   The `TimewarpProjectionDescription` class doesn't seem to be used at all.
*   What does this mean (documentation of `StatusFlag`)?

    > The values must be the same as in enum `StatusBits`.

*   Shouldn't `Session::pollController()` return the `InputState` instead of passing it by reference? Isn't the method chaining useless in this case?
*   The `HmdStatusFlag` enum is used only internally, shouldn't it be hidden from public API (putting it in `Implementation` namespace or in privates of `Session` class, e.g.)
*   Avoid repetition by renaming `Session::sessionStatus()` to just  `Session::status()`
*   `Session::ovrEyePoses()` returns `ovrPosef*` but the docs say that this is actually a two-element array. I suggest returning `Containers::StaticArrayView<2, ovrPosef>` instead.
*   `Session::headPoseState()`, `Session::calibratedOrigin()`, `Session::eyePoses()` and `Session::handPoseStates()` documentation is copy-pasted four times and doesn't match the code
*   Can we avoid `std::string` in the `Error` struct? What about just mirroring the original `char[512]`?

Squareys:
> There are some types still named Hmd but used in the Session class, is that okay?

Yes, that should be okay :)

[...]

PR #15, opened 2016-06-13
Cleanup OvrIntegration issues

Squareys:
Hi @mosra !

I finally got around to cleaning up #13... sorry it took me so long. And sorry for the mess, some of the things I fixed were rather sloppy.

Regards, Jonathan.

PS: Take your time with the review, I will need to compile check the example still anyway, and this shouldn't be urgent in any way.


src/Magnum/OvrIntegration/Compositor.cpp
```diff
@@ -108,8 +103,6 @@ LayerQuad& LayerQuad::setQuadSize(const Vector2& size) {
     return *this;
 }
 
-Compositor::Compositor() = default;
```
mosra: I would keep the constructor here, the class has heavy members (`std::vector`) and having the constructor inlined will increase compile time and binary size.

src/Magnum/OvrIntegration/Context.cpp
```diff
@@ -80,4 +80,15 @@ std::unique_ptr<Session> Context::createSession() {
     return std::unique_ptr<Session>(new Session(session));
 }
 
+Error Context::error() const {
+    ovrErrorInfo info;
+    ovr_GetLastErrorInfo(&info);
+
+    Error err;
+    err.type = ErrorType(info.Result);
+    strncpy(err.message, info.ErrorString, 512);
```
mosra: `std::strncpy()`?

mosra:
Thanks a lot! :+1: 

I still saw mentions of the OVR API instead of the wrapped API in the error enum, can you look there also?

Squareys:
@mosra I addressed the new issues (de-inlined the Compositor constructor, referenced OvrIntegration types in errors enum and added the `std::`). This could possibly be good to go now :)

Regards, Squareys.

Squareys:
@mosra Rebased and removed error enum values to know be OculusSDK 1.6.0 compliant!

I would be really happy, if this could get merged in the next 7 days, if that is possible somehow :)

src/Magnum/OvrIntegration/Session.cpp
```diff
@@ -37,6 +37,29 @@
...
+enum class HmdStatusFlag: UnsignedByte {
+    /**
+     * A mirror texture was created for the HMD and needs to be destroyed on
+     * destruction of the HMD
+     */
+    HasMirrorTexture = 1 << 0,
+
+    /** The HMD was created as a debug HMD (without real hardware) */
+    Debug = ovrHmdCap_DebugDevice /* 0x0010 */
+};
```
mosra: :+1: for the idea of having the enum definition only in the `*.cpp` file. A new thing for me! :)

mosra:
Merged. Thanks a lot!

Issue #11, opened 2015-12-29
BulletIntegration: btlDebugDraw implementation for Magnum.


Squareys:
Hello everybody!

While working with Bullet for a bit, I found this [btlDebugDraw](http://bulletphysics.org/Bullet/BulletFull/classbtIDebugDraw.html) class/interface which is supposed to be dervied with implementations for simple primitive drawing methods. It can then be attached to a `btCollisionWorld` or `btDynamicsWorld` for example, to provide fast and simple means of debug rendering.

I suggest this could be a great thing to have in the BulletIntegration library, since its code would most likely be common for all Magnum apps which use bullet.

(This is in no way urgent and if I have time, I might event implement this myself later next year. Just wanted to put that out there for discussion.)

Greetings, Squareys

mosra:
Hah, sorry, this somehow completely slipped through my attention...

Now I think this would be a much better approach than having those "shape converters", considering I'm about to deprecate/remove the half-~~assed~~implemented shapes library (https://github.com/mosra/magnum/issues/148).

Squareys:
We need to implement these methods of that interface:
 * virtual void setDebugMode (int debugMode)=0
 * virtual int getDebugMode () const =0
 * virtual void drawLine (const btVector3 &from, const btVector3 &to, const btVector3 &color)=0
 * virtual void reportErrorWarning (const char *warningString)=0
 * virtual void draw3dText (const btVector3 &location, const char *textString)=0
 * virtual void flushLines()

Everything else is already implemented using the `drawLine()` method. I suggest building a buffer with all the lines and then drawing that on `flushLines()`. That buffer needs to be updated every frame, so that's not optimal, but it's only for debug rendering anyway.

mosra:
Implemented via #21.

PR #21, opened 2016-10-02
Implement btIDebugDraw for Magnum

Squareys:
Hi @mosra !

As promised via gitter, here is the PR for the bullet physics visualization ( #11 ).

There is one `// TODO` in `draw3dText(...)`, you might want to look at how to specify/format it instead. I could also adapt the bullet example and whilst I'm at it remove the Shapes dependecy from it (since it will be deprecated) once you merged this.

Regards, Squareys.


Squareys:
> Coverage remained the same at 0.0%

Uh... I guess it's not set up for magnum-integration yet?

mosra:
No. If you remember, that's **STILL** because of this: https://github.com/travis-ci/apt-package-whitelist/issues/1190

I'm getting angry just seeing that issue.

The Bullet integration lib is built for OSX, but there I don't have the code coverage enabled. Maybe I should give up on waiting and build Bullet from source in this case.. :/

src/Magnum/BulletIntegration/DebugDraw.h
```diff
@@ -0,0 +1,177 @@
...
+            drawLine(pointOnB, pointOnB + normalOnB*distance, color);
+        }
+
+        /** @brief Print a warning */
+        virtual void reportErrorWarning(const char *warningString) override {
+            Warning() << "DebugDraw:" << warningString;
+        }
+
+        /** @brief Draw text */
+        virtual void draw3dText(const btVector3& CORRADE_UNUSED location, const char* CORRADE_UNUSED textString) override {
```
mosra: Just omit the argument name completely in case of unused parameters:

```cpp
void draw3dText(const btVector3&, const char*) override {
```

src/Magnum/BulletIntegration/DebugDraw.h
```diff
@@ -0,0 +1,177 @@
...
+            // TODO: Implement once debug text drawing is implemented.
+        }
+
+        /** @brief Flush lines to be drawn on screen */
+        virtual void flushLines() override;
```
mosra: I assume that all `override`n functions are called only from Bullet and not from user code. Could you make them all private to hide them from documentation?

The only public interface would then be `setDebugMode(Mode)` and `setViewProjectionMatrix()`.

Or is there a valid use case for calling them directly?

src/Magnum/BulletIntegration/DebugDraw.h
```diff
@@ -0,0 +1,177 @@
...
+
+        /** @brief Set view projection matrix used for rendering */
+        DebugDraw& setViewProjectionMatrix(const Matrix4& viewProj) {
```
mosra: According to naming scheme everywhere else I would name this function (and related variables/docs) `setTransformationProjectionMatrix()`.

Squareys:
> No. If you remember, that's STILL

Right, forgot about it, it's been a *year* since.

Any other comments on the PR? For once it is not WIP, was simple enough ;) (Not pushing, would be great to have this upstream, though.)

src/Magnum/BulletIntegration/DebugDraw.cpp
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    return debug << "BulletIntegration::DebugDraw::Mode::(invalid)";
+}
+
+void DebugDraw::drawLine(const btVector3& from, const btVector3& to, const btVector3& fromColor, const btVector3& toColor) {
+   _bufferData.push_back(Vector3(from));
+   _bufferData.push_back(Color3::from(fromColor));
+   _bufferData.push_back(Vector3(to));
+   _bufferData.push_back(Color3::from(toColor));
```
mosra: First: I realized that there is no way to create `Color` from an external representation. That I have to fix. Sorry.

Second: the only `::from()` I found is one that takes `T*`. And now I'm wondering, how the hell does this even compile? :D MSVC?!

src/Magnum/BulletIntegration/DebugDraw.h
mosra: Also, similarly to the constructor and destructor, I would move all the virtual function overrides into the `*.cpp` file. The compiler has to de-inline them anyway and by doing that explicitly you save some more compilation time when just including the header.

mosra:
Review done :)

The OSX build fails because of missing `Shaders` dependency. Can you update the CI files accordingly?

Regarding the text rendering -- yes, as you suggest, i think the better option long-term would be to use the "debug text" functionality once it's in. The TODO is fine like this. Or, maybe, put it in a Doxygen `@todo` block.

~~Until then, if you feel like doing it (just a suggestion!), even just copypasting the guts of the Text example and having the constructor take an *optional* `Text::AbstractFont* font` pointer with an *opened* font -- i.e. assert on it like this:~~

```cpp
CORRADE_INTERNAL_ASSERT(!font || font->isOpened());
```

~~If the user won't supply the parameter (the default), it will not render any text. If the user provides the text, it will render it.~~

Eh, scratch that, now realized that this will be kinda hard to do with this "immediate" approach of the interface, as there is no `flushText()` or anything similar. Would need to think about it a bit more.

src/Magnum/BulletIntegration/DebugDraw.cpp
Squareys: Yes, MSVC. I wondered, too. I will use the underlying float array, which is public.

src/Magnum/BulletIntegration/DebugDraw.cpp
```diff
@@ -0,0 +1,78 @@
...
+        _c(EnableCCD)
+        _c(DrawConstraints)
+        _c(DrawConstraintLimits)
+        _c(FastWirefram)
+        _c(DrawNormals)
+        _c(DrawFrames)
+        #undef _c
+    }
+
+    return debug << "BulletIntegration::DebugDraw::Mode::(invalid)";
```
Squareys: Alright, will do 👍

src/Magnum/BulletIntegration/DebugDraw.h
```diff
@@ -0,0 +1,177 @@
...
+
+#include <Corrade/Containers/ArrayView.h>
+
+#include <Magnum/Buffer.h>
+#include <Magnum/DebugOutput.h>
+#include <Magnum/Magnum.h>
+#include <Magnum/Mesh.h>
+#include <Magnum/Shaders/VertexColor.h>
+
+#include <bullet/LinearMath/btIDebugDraw.h>
```
Squareys: Ok, did not know that, will do.

src/Magnum/BulletIntegration/DebugDraw.h
```diff
@@ -0,0 +1,177 @@
...
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+/** @file
+ * @brief Implementation of @ref btIDebugDraw for physics visualization in bullet
```
Squareys: Ok, hence: Remove the `@ref` and have "Bullet" with capital `B`.

src/Magnum/BulletIntegration/DebugDraw.h
```diff
@@ -0,0 +1,177 @@
...
+            DisableBulletLCP = DBG_DisableBulletLCP,    /**< Disable Bullet LCP */
+            EnableCCD = DBG_EnableCCD,                  /**< Enable CCD */
+            DrawConstraints = DBG_DrawConstraints,      /**< Draw constaints */
+            DrawConstraintLimits = DBG_DrawConstraintLimits, /**< Draw constraint limits */
+            FastWirefram = DBG_FastWireframe,           /**< Draw fast wireframes */
+            DrawNormals = DBG_DrawNormals,              /**< Draw normals */
+            DrawFrames = DBG_DrawFrames,                /**< Draw frames */
+        };
+
+        typedef Containers::EnumSet<Mode> Modes;
```
Squareys: `/** @brief Debug draw mode flags */` ? :)

src/Magnum/BulletIntegration/DebugDraw.h
```diff
@@ -0,0 +1,177 @@
...
+         *
+         * Sets up @ref Shaders::VertexColor3D, @ref Buffer and @ref Mesh for physics debug rendering.
+         */
+        DebugDraw(UnsignedInt initialBufferCapacity=0):
+            btIDebugDraw(),
+            _mesh{MeshPrimitive::Lines}
+        {
+            _mesh.addVertexBuffer(_buffer, 0, Shaders::VertexColor3D::Position{}, Shaders::VertexColor3D::Color{});
+            _bufferData.reserve(initialBufferCapacity*4);
+        }
```
Squareys: So... instead of providing a default value for `initialCapacity` parameter, right?

src/Magnum/BulletIntegration/DebugDraw.h
mosra: Yes, and "flag" instead of "enum" for the `Mode` enum.

src/Magnum/BulletIntegration/DebugDraw.cpp
mosra: Commit b033af3b593f536397628c44c9ff3c6ce6ed3628 fixes the Color conversion so you can now make it the same way as with `Vector3`.

src/Magnum/BulletIntegration/DebugDraw.h
Squareys: 1. [x] done
2. [ ] done

src/Magnum/BulletIntegration/DebugDraw.h
Squareys: But some day you may want to implement it and then you would need to look up what the paramters mean? :)

src/Magnum/BulletIntegration/DebugDraw.h
```diff
@@ -0,0 +1,177 @@
...
+
+        /** @brief Set debug mode */
+        virtual void setDebugMode(int debugMode) override {
+            _debugMode = Mode(debugMode);
+        }
+
+        /** @overload */
+        virtual void setDebugMode(Modes debugMode) {
```
Squareys: Nope, totally right, found that just now myself also 👍

src/Magnum/BulletIntegration/DebugDraw.h
```diff
@@ -0,0 +1,177 @@
...
+        DebugDraw(UnsignedInt initialBufferCapacity=0):
+            btIDebugDraw(),
+            _mesh{MeshPrimitive::Lines}
+        {
+            _mesh.addVertexBuffer(_buffer, 0, Shaders::VertexColor3D::Position{}, Shaders::VertexColor3D::Color{});
+            _bufferData.reserve(initialBufferCapacity*4);
+        }
+
+        /** @brief Set debug mode */
+        virtual void setDebugMode(int debugMode) override {
```
Squareys: - [x] done

src/Magnum/BulletIntegration/DebugDraw.h
Squareys: - [x] done

src/Magnum/BulletIntegration/DebugDraw.h
Squareys: - [x] done

src/Magnum/BulletIntegration/DebugDraw.h
Squareys: - [x] done

Was not aware of that 👍

src/Magnum/BulletIntegration/DebugDraw.h
Squareys: - [x] done

src/Magnum/BulletIntegration/DebugDraw.h
mosra: Good point :)

Assuming the declaration is in header and implementation in the source file, the header can retain the parameter names and the implementation can have them omitted.

mosra:
Cherry-picked as f8eb01d66d44569b4ffb8000cc51922dbb28a43e, 28eb3c7d0c02d7d79fa7f2d5d1d5c03ff93bb96c and de3b54e3f10662d1b57db73fafa6e2f2ae7bf33b. On Linux the test failed the same way as on OSX, so I added an explicit cast to `UnsignedInt` before casting to `void*` to avoid sign extension.

Squareys:
What's with https://github.com/mosra/magnum-integration/pull/21/commits/a08be624805ffe37df27f39f0ea6d114c06e4ce5?

mosra:
Into both, actually :)

I'm leaving the cleanup commits only in case they are not by the same author (so people can see the changes others made to their code).

PR #22, opened 2016-10-11
Fix MotionState

Squareys:
Hi @mosra !

Here is the mentioned pullrequest to finally fix the behaviour of `BulletIntegration::MotionState` with parents in the SceneGraph.

While I was at it, I added more conversions, a test and did the `@todoc` in MotionState.h

Regards, Squareys.

### DIFF
--- a/src/Magnum/BulletIntegration/DebugDraw.h
+++ b/src/Magnum/BulletIntegration/DebugDraw.h
@@ -32,6 +32,7 @@
 
 #include <vector>
 #include <Corrade/Containers/ArrayView.h>
+#include <Corrade/Utility/Macros.h>
 #include <Magnum/Buffer.h>
 #include <Magnum/DebugOutput.h>
 #include <Magnum/Magnum.h>
--- a/src/Magnum/BulletIntegration/MotionState.cpp
+++ b/src/Magnum/BulletIntegration/MotionState.cpp
@@ -26,27 +26,50 @@
 
 #include "MotionState.h"
 
+#include <Magnum/Math/Angle.h>
 #include <Magnum/Math/Matrix4.h>
 
 #include "Magnum/BulletIntegration/Integration.h"
 
+
 namespace Magnum { namespace BulletIntegration {
 
+using namespace Math::Literals;
+
 void MotionState::getWorldTransform(btTransform& worldTrans) const {
-    const Matrix4 transformation = object().transformationMatrix();
-    worldTrans.setOrigin(btVector3(transformation.translation()));
-    worldTrans.setBasis(btMatrix3x3(transformation.rotationScaling()));
+    /* Bullet transforms are always world space */
+    worldTrans = btTransform{object().absoluteTransformationMatrix()};
 }
 
 void MotionState::setWorldTransform(const btTransform& worldTrans) {
-    btVector3 bPosition = worldTrans.getOrigin();
-    btVector3 bAxis = worldTrans.getRotation().getAxis();
-    Rad bRotation(worldTrans.getRotation().getAngle());
+    auto parent = object().parent();
+    if(parent) {
+        /* Get local transform Bullet world transform by removing parent transform */
+        const Matrix4 parentTransform = parent->absoluteTransformationMatrix();
 
-    /** @todo Verify that all objects have common parent */
-    transformation.resetTransformation()
-        .rotate(bRotation, Vector3(bAxis).normalized())
-        .translate(Vector3(bPosition));
+        CORRADE_ASSERT(parentTransform.isRigidTransformation(), "MotionState::setWorldTransform(): parent of object with MotionState cannot have scaling", );
+        const Quaternion parentRotation = Quaternion::fromMatrix(parentTransform.rotationScaling());
+
+        const Quaternion rotation = (parentRotation.inverted()*Quaternion{worldTrans.getRotation()})
+                                    .normalized(); /* Avoid floating point drift */
+        const Vector3 translation = Vector3{worldTrans.getOrigin()} - parentTransform.translation();
+
+        /* Apply calculated transformation to object */
+        _transformation.resetTransformation();
+
+        const Rad angle = rotation.angle();
+        if(angle != 0.0_radf) {
+            /* If angle is 0, the axis would be a NaN vector */
+            _transformation.rotate(angle, rotation.axis());
+        }
+
+        _transformation.translate(translation);
+    } else {
+        const Quaternion rotation = Quaternion{worldTrans.getRotation()};
+        _transformation.resetTransformation()
+                       .rotate(rotation.angle(), rotation.axis())
+                       .translate(Vector3{worldTrans.getOrigin()});
+    }
 }
 
 }}
--- a/src/Magnum/BulletIntegration/MotionState.h
+++ b/src/Magnum/BulletIntegration/MotionState.h
@@ -31,7 +31,7 @@
  */
 
 #include <LinearMath/btMotionState.h>
-#include <Magnum/SceneGraph/Object.h>
+#include <Magnum/SceneGraph/AbstractFeature.h>
 #include <Magnum/SceneGraph/AbstractTranslationRotation3D.h>
 
 #include "Magnum/BulletIntegration/visibility.h"
--- a/src/Magnum/BulletIntegration/MotionState.h
+++ b/src/Magnum/BulletIntegration/MotionState.h
@@ -43,7 +43,31 @@ namespace Magnum { namespace BulletIntegration {
 
 Encapsulates `btMotionState` as @ref SceneGraph feature.
 
-@todoc Usage...
+# Usage
+
+Common usage is to either create a `btRigidBody` to share transformation with
+a @ref SceneGraph::Object by passing the motion state in its constructor:
+
+@code
+    SceneGraph::Object<SceneGraph::MatrixTransformation3D> object;
+    btRigidBody rigidBody{mass, &(new MotionState{object}).btMotionState(), collisionShape};
+
+    // rigidBody will be affected when changing the transform of object and
+    // object will be affected when transformation of rigidBody is changed.
+@endcode
+
+Or setting the motion state afterwards:
+
+@code
+    SceneGraph::Object<SceneGraph::MatrixTransformation3D> object;
+    btRigidBody rigidBody{...};
+    rigidBody.setMotionState(&(new MotionState{object}).btMotionState());
+@endcode
+
+Note that changes to a rigidBody using `btRigidBody::setWorldTransform()` may
+only update the motion state of non-static objects and while
+`btDynamicsWorld::stepSimulation()` is called.
+
 */
 class MAGNUM_BULLETINTEGRATION_EXPORT MotionState: public SceneGraph::AbstractBasicFeature3D<btScalar>, private btMotionState {
     public:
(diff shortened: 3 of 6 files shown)


src/Magnum/BulletIntegration/Test/CMakeLists.txt
```diff
@@ -24,5 +25,5 @@
 #   DEALINGS IN THE SOFTWARE.
 #
 
-corrade_add_test(BulletIntegrationTest IntegrationTest.cpp LIBRARIES MagnumBulletIntegration)
+corrade_add_test(BulletIntegrationTest IntegrationTest.cpp LIBRARIES MagnumBulletIntegration ${BULLET_LIBRARIES})
```
mosra: This shouldn't be needed -- the bullet libraries are already a dependency of `MagnumBulletIntegration`.

src/Magnum/BulletIntegration/Test/IntegrationTest.cpp
```diff
@@ -106,6 +123,46 @@ void IntegrationTest::quaternion() {
...
+          .rotateY(25.0_degf)
+          .rotateZ(15.0_degf);
+    child.translate(Vector3{30.0f, 20.0f, 10.0f})
+          .rotateZ(32.0_degf)
+          .rotateY(22.0_degf)
+          .rotateX(12.0_degf);
+
+    /* Setup rigid body which will receive the motion state */
+    MotionState motionState{child};
+    btSphereShape collisionShape{0.0};
```
mosra: This should be `0.0f` -- some compilers are taking type narrowing very seriously. I think you should get at least a warning from the compiler.

Btw., how is it with `btScalar`? Is it always `float` or can it be switched to `double` at compile time?

src/Magnum/BulletIntegration/Test/IntegrationTest.cpp
```diff
@@ -29,14 +29,29 @@
...
+
+#include <Magnum/SceneGraph/MatrixTransformation3D.h>
+#include <Magnum/SceneGraph/Object.h>
+#include <Magnum/SceneGraph/Scene.h>
+
 #include "Magnum/BulletIntegration/Integration.h"
 #include "Magnum/BulletIntegration/DebugDraw.h"
+#include "Magnum/BulletIntegration/MotionState.h"
+
+#include <BulletDynamics/btBulletDynamicsCommon.h>
```
mosra: (Okay, I'm just being annoying here. Sorry.)

I would not separate the includes with empty lines. Just two groups -- one with `<>` and the second with `""`.

EDIT: I saw you fixed that in a later commit -- could you squash those two together?

src/Magnum/BulletIntegration/MotionState.h
```diff
@@ -43,7 +43,31 @@ namespace Magnum { namespace BulletIntegration {
 
 Encapsulates `btMotionState` as @ref SceneGraph feature.
 
-@todoc Usage...
+# Usage
```
mosra: This should be `## Usage`, otherwise the header gets way too big in Doxygen output.

src/Magnum/BulletIntegration/MotionState.h
```diff
@@ -43,7 +43,31 @@ namespace Magnum { namespace BulletIntegration {
...
+Common usage is to either create a `btRigidBody` to share transformation with
+a @ref SceneGraph::Object by passing the motion state in its constructor:
+
+@code
+    SceneGraph::Object<SceneGraph::MatrixTransformation3D> object;
+    btRigidBody rigidBody{mass, &(new MotionState{object}).btMotionState(), collisionShape};
+
+    // rigidBody will be affected when changing the transform of object and
+    // object will be affected when transformation of rigidBody is changed.
+@endcode
```
mosra: Please don't indent stuff inside `@code` -- it will get indented also in the documentation, wasting precious horizontal space.

src/Magnum/BulletIntegration/DebugDraw.h
```diff
@@ -32,6 +32,7 @@
 
 #include <vector>
 #include <Corrade/Containers/ArrayView.h>
+#include <Corrade/Utility/Macros.h>
```
mosra: Oh. I guess this slipped through in the previous PR... The parameter names don't need to be marked with `CORRADE_UNUSED` in function *declaration* in the header -- it will not generate any compiler warning. So then this include is also not needed.

src/Magnum/BulletIntegration/MotionState.cpp
```diff
@@ -26,27 +26,50 @@
 
 #include "MotionState.h"
 
+#include <Magnum/Math/Angle.h>
```
mosra: Hmm. Isn't this included implicitly by the `Matrix4.h`?

src/Magnum/BulletIntegration/MotionState.cpp
```diff
@@ -26,27 +26,50 @@
...
 #include <Magnum/Math/Matrix4.h>
 
 #include "Magnum/BulletIntegration/Integration.h"
 
+
```
mosra: Why? ;)

src/Magnum/BulletIntegration/Test/CMakeLists.txt
```diff
@@ -25,5 +25,5 @@
 #   DEALINGS IN THE SOFTWARE.
 #
 
-corrade_add_test(BulletIntegrationTest IntegrationTest.cpp LIBRARIES MagnumBulletIntegration ${BULLET_LIBRARIES})
+corrade_add_test(BulletIntegrationTest IntegrationTest.cpp LIBRARIES MagnumBulletIntegration)
```
mosra: Ah. Okay. That makes my previous comment on this line obsolete :) Would it be possible to discard the change from the commits altogether? ;)

src/Magnum/BulletIntegration/MotionState.h
```diff
@@ -69,25 +69,27 @@ only update the motion state of non-static objects and while
...
+        ~MotionState() = default;
+
         /** @brief Motion state */
         btMotionState& btMotionState() { return *this; }
 
     private:
-        void MAGNUM_BULLETINTEGRATION_LOCAL getWorldTransform(btTransform& worldTrans) const override;
-        void MAGNUM_BULLETINTEGRATION_LOCAL setWorldTransform(const btTransform& worldTrans) override;
+        void MAGNUM_BULLETINTEGRATION_EXPORT getWorldTransform(btTransform& worldTrans) const override;
+        void MAGNUM_BULLETINTEGRATION_EXPORT setWorldTransform(const btTransform& worldTrans) override;
```
mosra: I saw the OSX CI fail on some linker error. I guess this is probably related to the above changes of the EXPORT/LOCAL macros. I think reverting those and defaulting the destructor inside the `*.cpp` file (instead of having it inline in the header) might help.

Overriden virtuals don't need to be exported (unless you are calling them directly, which you can't, because they are private).

src/Magnum/BulletIntegration/MotionState.cpp
```diff
@@ -26,27 +26,50 @@
...
-    btVector3 bPosition = worldTrans.getOrigin();
-    btVector3 bAxis = worldTrans.getRotation().getAxis();
-    Rad bRotation(worldTrans.getRotation().getAngle());
-
-    /** @todo Verify that all objects have common parent */
-    transformation.resetTransformation()
-        .rotate(bRotation, Vector3(bAxis).normalized())
-        .translate(Vector3(bPosition));
+    auto parent = object().parent();
+    if(parent) {
```
mosra: This will be almost always true, no? Because even the scene is a parent.

src/Magnum/BulletIntegration/MotionState.cpp
```diff
@@ -26,27 +26,50 @@
```
mosra: There's some noise in the commit message ;)

src/Magnum/BulletIntegration/MotionState.cpp
```diff
@@ -26,27 +26,50 @@
...
+        const Quaternion rotation = (parentRotation.inverted()*Quaternion{worldTrans.getRotation()})
+                                    .normalized(); /* Avoid floating point drift */
+        const Vector3 translation = Vector3{worldTrans.getOrigin()} - parentTransform.translation();
+
+        /* Apply calculated transformation to object */
+        _transformation.resetTransformation();
+
+        const Rad angle = rotation.angle();
+        if(angle != 0.0_radf) {
+            /* If angle is 0, the axis would be a NaN vector */
```
mosra: Added myself a TODO: https://github.com/mosra/magnum/projects/5#card-472678

src/Magnum/BulletIntegration/MotionState.cpp
```diff
@@ -26,27 +26,50 @@
...
+    if(parent) {
+        /* Get local transform Bullet world transform by removing parent transform */
+        const Matrix4 parentTransform = parent->absoluteTransformationMatrix();
+
+        CORRADE_ASSERT(parentTransform.isRigidTransformation(), "MotionState::setWorldTransform(): parent of object with MotionState cannot have scaling", );
+        const Quaternion parentRotation = Quaternion::fromMatrix(parentTransform.rotationScaling());
+
+        const Quaternion rotation = (parentRotation.inverted()*Quaternion{worldTrans.getRotation()})
+                                    .normalized(); /* Avoid floating point drift */
+        const Vector3 translation = Vector3{worldTrans.getOrigin()} - parentTransform.translation();
```
mosra: Hmm. All the inversions, `sqrt()` when converting the matrix to quaternion and renormalizations are quite costly operations and I'm not sure how this would perform with 100s of objects.

But on the other hand I also don't see any better way how to handle this if you really want to cover the general case where every object can have a different parent but you still want them all in a single physics world.

Things to consider:

- there is transformation caching that could be maybe used for this purpose ([docs](http://mosra.cz/blog/magnum-doc/scenegraph.html#scenegraph-features-caching)) assuming that the parent transformation doesn't change that much (otherwise it would be just more overhead)
- there's also a way to get absolute transformations of a bunch of objects at once (not just one) using [SceneGraph::Object](http://mosra.cz/blog/magnum-doc/classMagnum_1_1SceneGraph_1_1Object.html#a87a8cd6facd28a091523f5b35c6e86a9) and that would also save some calculation time when more objects share the same parent. Does Bullet have some way to batch update transformations of more objects where this could be used? Unfortunately there's no way to get inverse absolute transformations (yet), I would need to add that.

mosra:
I completely forgot that this has been a thing :) Thanks a lot, it now helped me when fixing the DLL linker errors on Windows. I cherry-picked most of the things in 3a440409d9d618b19f221db07a4e472cc5774779, bff3f7ccf861a085308a64e66bf5b9b8be0bc3ce, ba5b1a7b91dbb4c0e0fca7e08a7369fdf77885dc, d6941bf49f050738470f898063c5d3602582b85a, c45d5a2b160a20ece4c0a377f5f73f5be7a1d311 and a971a8855b2a22adb03705ba0ef2a52183ed333b and thanks to the new test the code coverage is now doubled :)

Just for the record, the "Fix local transforms from bullet world transforms" commit is not merged as we finally agreed that this was too much performance impact for too little benefit. Just aside: isn't this behavior practically the same in UE4? IIRC they also just disconnect a physics object from all the parents when it starts to be simulated.

Squareys:
> behavior practically the same in UE4?

While, yes, it is disconnected, you can reconnect it to have it act in weird unexpected ways (simulated, but also getting relative transforms of the parent applied to it, but I think also depends on update order, but not sure...).

As long as the documentation is clear about it, I'm fine with the parent of the object being ignored :)

PR #24, opened 2017-03-20
Allow Bullet integration in the face of -DBUILD_DEPRECATED=OFF

alicemargatroid:
Bullet Integration should probably only use non-deprecated functions / methods.


Squareys:
Oh, good catch!

mosra:
Thanks a lot for this! 

I have quite a huge amount of deprecation warnings in the plugins repo as well. Had to push some important features into the engine to make a release at my company possible and due to an extreme lack of time I was not able to tie all the loose ends early enough :(

As I said in your other PR in the corrade repo, I'm now adding explicit (and [scheduled](https://blog.travis-ci.com/2016-12-06-the-crons-are-here)) build jobs on the CI with `BUILD_DEPRECATED` set to `OFF` to prevent further accidents like this.

PR #26, opened 2017-09-14
OvrIntegration: Adapt to API changes in Oculus SDK 1.17.0

Squareys:
Hi @mosra !

Just came across this, this is not urgent, merge whenever you find the time.

Cheers, Jonathan

### DIFF
--- a/package/ci/appveyor.yml
+++ b/package/ci/appveyor.yml
@@ -62,8 +62,8 @@ install:
 - IF "%TARGET%" == "desktop" IF "%COMPILER%" == "mingw" pip install codecov
 
 # OVR SDK
-- IF "%TARGET%" == "desktop" IF "%COMPILER%" == "msvc" IF NOT EXIST %APPVEYOR_BUILD_FOLDER%\ovr_sdk_win_1.4.0_public.zip appveyor DownloadFile https://static.oculus.com/sdk-downloads/1.4.0/Public/1463687419/ovr_sdk_win_1.4.0_public.zip
-- IF "%TARGET%" == "desktop" IF "%COMPILER%" == "msvc" 7z x ovr_sdk_win_1.4.0_public.zip -o%APPVEYOR_BUILD_FOLDER%\deps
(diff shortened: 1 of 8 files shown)


mosra:
Thank you!

One problem I see immediately (maybe you remember): there was no easy way to get the latest SDK from Oculus website, so the CI is stuck on something very outdated. Any idea how to solve this (besides hosting it privately)?

Squareys:
@mosra Yup, see last commit. I wonder if the access token expires sometime... otherwise we can refresh it and hope for the cache staying valid? :P

Squareys:
Well, alright, it seems to download something else it seems...

mosra:
Huh, this is still a thing? :D

Squareys:
@mosra Yes, we didn't merge it, because we didn't manage to get the ovr sdk downloaded by the CI.

Squareys:
Actually, now that I think of it, we wanted to try whether the download key expires somehow. Also, it looks like I need to add some lines to make it build the Debug libraries of the SDK.

Squareys:
Summarizing offline discussions:

Solution to Oculus SDK not being automatically downloadable anymore is to host some minimal archive required for the build and use that. To generate it, I added a python script with 869ca71.
This is useful anyway, since LibOVR links static CRT by default and therefore is not compatible with all other magnum libraries (which link dynamic CRT by default) and we therefore need to patch the VS projects.

This will be merged once mosra finds the time after some important bigger changes in the main repository.

mosra:
Random idea since I don't have time to set up `ci.magnum.graphics` yet (and won't have for the next x weeks): does the SDK provide some version define that could be used to make it temporarily work both with the currently still downloadable SDK and the latest one? Yes, it's ugly and all, but better than have this totally non-compilable with latest SDK -- I may have a need for a working version soon.

Squareys:
>  it took me almost 12 months to merge

Better late than never 😆 Since nobody complained yet... :D

Thank you for merging! 🎉

PR #27, opened 2017-10-01
BulletIntegration: Remove DebugOutput include from DebugDraw

Squareys:
Hi @mosra !

This include prevents build with emscripten (as it is not available there) and is not required.

Cheers, Jonathan


mosra:
Thank you!

Squareys:
Thank you for the quick merge! 🎉

PR #28, opened 2018-01-23
Dart Integration

costashatz:
This PR adds an integration to the [DART simulator](http://dartsim.github.io/). In short:

- Provides `DARTObject` for encapsulating dart's `BodyNode` or `ShapeNode` as a `SceneGraph` 3D feature for transformations keeping
- Provides `DARTSkeleton` for easy bookkeeping of dart's `Skeleton` (i.e., Multi-Body entity)
- Provides `ConvertShapeNode` that reads the shape of a `ShapeNode` (or a `DARTObject` that contains a `ShapeNode`) and returns a struct (`VisualData`) that has useful `Magnum::Trade` that can be used to render the shapes
- Unit tests to test functionality for simple objects, meshes and textured meshes
- An example using this DartIntegration is already almost done (see [here](https://github.com/costashatz/magnum-examples/tree/dart-example)).

Illustration:
ezgif-5-e1da75dd98

@mosra let me know of any changes you would like me to do!


src/Magnum/DartIntegration/DartObject.h
```diff
@@ -0,0 +1,137 @@
...
+
+        /** @brief get transformation from Dart */
+        DartObject& update();
+
+        /** @brief ShapeNode */
+        dart::dynamics::ShapeNode* shapeNode();
+        /** @brief BodyNode */
+        dart::dynamics::BodyNode* bodyNode();
+
+    protected:
```
mosra: Does this need to be protected? I would make it `private`.

src/Magnum/DartIntegration/DartSkeleton.cpp
```diff
@@ -0,0 +1,63 @@
...
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+#include <Magnum/DartIntegration/DartSkeleton.h>
+
+namespace Magnum { namespace DartIntegration {
+
+DartSkeleton& DartSkeleton::updateObjects()
+{
```
mosra: Coding style: please put the opening brackets on the same line, thank you :)

Similarly in a bunch other places.

src/Magnum/DartIntegration/DartSkeleton.h
```diff
@@ -0,0 +1,140 @@
...
+
+        /** @brief get all objects */
+        std::vector<DartObject*> objects();
+        /** @brief get all objects associated with `ShapeNode`s */
+        std::vector<DartObject*> shapeObjects();
+        /** @brief get all objects associated with `BodyNode`s */
+        std::vector<DartObject*> bodyObjects();
+
+    protected:
+        std::vector<DartObject*> _objects;
```
mosra: Looking at the code, the `DartObject*` pointers are never `nullptr`.

Assuming this class is owner of the `DartObject` instances, could the `_objects` be turned into `std::vector<std::unique_ptr<DartObject>>` and the functions above return `std::vector<std::reference_wrapper<DartObject>>`?

src/Magnum/DartIntegration/DartSkeleton.h
```diff
@@ -0,0 +1,140 @@
...
+        DartSkeleton& updateObjects();
+
+        /** @brief get all objects */
+        std::vector<DartObject*> objects();
+        /** @brief get all objects associated with `ShapeNode`s */
+        std::vector<DartObject*> shapeObjects();
+        /** @brief get all objects associated with `BodyNode`s */
+        std::vector<DartObject*> bodyObjects();
+
+    protected:
```
mosra: Again, does this need to be protected? I would make it `private`.

src/Magnum/DartIntegration/DartSkeleton.h
```diff
@@ -0,0 +1,140 @@
...
+        std::vector<DartObject*> _objects;
+
+        template <typename T> void _parseBodyNodeRecursive(T* parent, dart::dynamics::BodyNode* bn) {
+            if (!bn || !parent)
+                return;
```
mosra: As far as I can see, `nullptr` is never passed to `parent`. Could that be made into a reference instead? Is the same possible for the `bn` parameter? Would make the code a bit clearer.

src/Magnum/DartIntegration/Test/DartSkeletonTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+const double default_rest_position = 0.0;
+const double delta_rest_position = 10.0 * M_PI / 180.0;
+
+const double default_stiffness = 0.0;
+const double delta_stiffness = 10;
+
+const double default_damping = 5.0;
+const double delta_damping = 1.0;
+
+inline void setGeometry(const dart::dynamics::BodyNodePtr& bn) {
```
mosra: Could you wrap all the variables above and these `inline` functions into an anonymous namespace? The functions then don't need to be `inline` anymore (the compiler warning related to that will go away).

src/Magnum/DartIntegration/Test/DartSkeletonTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+const double default_height = 1.0; // m
+const double default_width = 0.2; // m
+const double default_depth = 0.2; // m
+
+const double default_torque = 15.0; // N-m
+const double default_force = 15.0; // N
+const int default_countdown = 200; // Number of timesteps for applying force
+
+const double default_rest_position = 0.0;
+const double delta_rest_position = 10.0 * M_PI / 180.0;
```
mosra: Wanted to suggest using `constexpr Rad delta_rest_position = 10.0_degf` here, but the variable isn't used anywhere, if I see correctly.

src/Magnum/DartIntegration/Test/DartSkeletonTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+#include <Magnum/Trade/MeshData3D.h>
+#include <Magnum/Trade/PhongMaterialData.h>
+#include <Magnum/Trade/TextureData.h>
+
+#include <Magnum/DartIntegration/ConvertShapeNode.h>
+#include <Magnum/DartIntegration/DartSkeleton.h>
+
+#include <Magnum/DartIntegration/Test/configure.h>
+
+const double default_height = 1.0; // m
```
mosra: Coding style: since these are constants, can they be named CamelCase and `constexpr`? So e.g.

```cpp
constexpr Double DefaultHeight = 1.0;
```

Thank you :)

src/Magnum/DartIntegration/Test/DartSkeletonTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+    typedef SceneGraph::Scene<SceneGraph::MatrixTransformation3D> Scene3D;
+
+}
+
+struct DartSkeletonTest : TestSuite::Tester {
+    explicit DartSkeletonTest();
+
+    void test();
+    void test_urdf();
+    void test_texture();
```
mosra: Coding style: no underscores, please :) Also you can drop the `test` prefix and have just `urdf()` and `texture()` there.

src/Magnum/DartIntegration/Test/DartSkeletonTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+};
+
+DartSkeletonTest::DartSkeletonTest() {
+    addTests({&DartSkeletonTest::test});
+    addTests({&DartSkeletonTest::test_urdf});
+    addTests({&DartSkeletonTest::test_texture});
```
mosra: It's possible to write the above like this (with the naming adapted already):

```cpp
addTests({&DartSkeletonTest::test,
          &DartSkeletonTest::urdf,
          &DartSkeletonTest::texture});
```

src/Magnum/DartIntegration/Test/DartSkeletonTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+    Scene3D scene;
+    Object3D* obj = new Object3D{&scene};
+
+    DartSkeleton* skel = new DartSkeleton{obj, pendulum};
+
+    // Set the initial joint positions so that the pendulum
+    // starts to swing right away
+    pendulum->getDof(1)->setPosition(120. * M_PI / 180.0);
+    pendulum->getDof(2)->setPosition(20. * M_PI / 180.0);
+    pendulum->getDof(3)->setPosition(-50. * M_PI / 180.0);
```
mosra: There's the `Constants::pi()` constant if you want, you could also do `Float(Rad(120.0_degf))`, but that might look scary and I'm not forcing you to do that ;)

src/Magnum/DartIntegration/Test/DartSkeletonTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+    auto trans = bn->getShapeNodesWith<dart::dynamics::VisualAspect>().back()->getTransform();
+
+    auto R = Eigen::AngleAxisd(trans.linear());
+    auto axis = R.axis();
+    auto T = trans.translation();
+
+    // Convert it to axis-angle representation
+    Math::Vector3<Float> t = {Float(T[0]), Float(T[1]), Float(T[2])};
+    Math::Vector3<Float> u = {Float(axis(0)), Float(axis(1)), Float(axis(2))};
+    Float theta = Float(R.angle());
```
mosra: Similarly to my comment in `DartObject.cpp` above -- can the types be explicit instead of `auto`? Also, the math type conversion can be written in a shorter way.

> axis-angle representation

This caught my attention here, but applies to `DartObject,cpp` as well. Wouldn't there be some simpler, more direct way? (matrices? dual quats?) If I remember correctly, converting to axis/angle involves some nasty square roots and acos and so it's not exactly fast. I saw Eigen for the first time today, so I have no idea in what representation the `Isometry3d` type stores the data :D

src/Magnum/DartIntegration/Test/DartSkeletonTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+    for (size_t i = 0; i < skel->objects().size(); i++) {
+        if (!skel->objects()[i]->shapeNode())
+            continue;
+        auto shape = skel->objects()[i]->shapeNode()->getShape();
+        CORRADE_VERIFY(shape->getType() == "MeshShape");
+        auto mydata = convertShapeNode(skel->objects()[i]->shapeNode());
+        CORRADE_VERIFY(mydata);
+        CORRADE_VERIFY((*mydata)._meshData->positions(0).size() == pts[j]);
+        Vector3 color{0.6f, 0.6f, 0.6f};
+        CORRADE_VERIFY((*mydata)._matData->diffuseColor() == color);
```
mosra: Using `CORRADE_COMPARE((*mydata)._matData->diffuseColor(), color)` instead of `CORRADE_VERIFY()` improves the diagnostic output better in case of a failure (i.e., it prints both values instead of just saying that comparison failed). That is great to have when debugging test failures on the CI, for example. Would be great if you could use it in all cases where it's comparing more than just a boolean :)

mosra:
Oh, one more thing: what's the size of the test data, roughly? GitHub doesn't want to tell me :)

costashatz:
> Oh, one more thing: what's the size of the test data, roughly? GitHub doesn't want to tell me :)

I removed the big files (most probably they are in the git memory, but I guess we can remove them). So, the total file size of the test data at the moment is roughly 450KB!

mosra:
450 kB? for all that? That's cool. Much less than data for some audio importer tests :D

costashatz:
> 450 kB? for all that? That's cool. Much less than data for some audio importer tests :D

There's nothing really. The STL files occupy most of the space and then I borrowed the textured collada file from the `AssimpImporter` test. The URDF files are just text files and we have a few degrees of freedom for the arm..!

src/Magnum/DartIntegration/ConvertShapeNode.cpp
```diff
@@ -0,0 +1,193 @@
...
+    Eigen::Vector4d col = shapeNode->getVisualAspect()->getRGBA();
+
+    // default shininess to 80.f
+    data._matData = new Trade::PhongMaterialData{Trade::PhongMaterialData::Flags{}, 80.f};
+    data._matData->ambientColor() = Color3{Float(col(0)), Float(col(1)), Float(col(2))};
+    data._matData->diffuseColor() = data._matData->ambientColor();
+    data._matData->specularColor() = data._matData->ambientColor();
+
+    if (shape->getType() == "BoxShape") {
+        auto boxShape = std::dynamic_pointer_cast<dart::dynamics::BoxShape>(shape);
```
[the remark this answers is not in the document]
costashatz: I think `dynamic_pointer_cast` is exactly what we need (and what the Dart developers suggest to use in this case), because a run-time check is performed: https://stackoverflow.com/questions/28002/regular-cast-vs-static-cast-vs-dynamic-cast

Btw, I could/should check if the pointer is valid!

src/Magnum/DartIntegration/ConvertShapeNode.h
```diff
@@ -0,0 +1,82 @@
...
+*/
+Corrade::Containers::Optional<VisualData> MAGNUM_DARTINTEGRATION_EXPORT convertShapeNode(dart::dynamics::ShapeNode* shapeNode);
+
+/**
+@brief Convert DartObject to @ref VisualData that can be processed by Magnum
+
+See @ref convertShape(dart::dynamics::ShapeNode* shapeNode)
+for more information.
+*/
+Corrade::Containers::Optional<VisualData> MAGNUM_DARTINTEGRATION_EXPORT convertShapeNode(DartObject* object);
```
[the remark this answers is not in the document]
costashatz: > Could these two functions take a reference instead of a pointer? So the nullptr check responsibility is moved to the user of the API.

I think they can yes. But why is it better to move the nullptr check to the user?

src/Magnum/DartIntegration/DartSkeleton.h
costashatz: > Assuming this class is owner of the DartObject instances, could the _objects be turned into std::vector<std::unique_ptr<DartObject>> and the functions above return std::vector<std::reference_wrapper<DartObject>>?

Although I agree with you, wouldn't the `std::reference_wrapper` make the access to the objects I bit more tedious (one will have to use `.get()` etc)?

src/Magnum/DartIntegration/Test/DartSkeletonTest.cpp
costashatz: > This caught my attention here, but applies to DartObject,cpp as well. Wouldn't there be some simpler, more direct way? (matrices? dual quats?) If I remember correctly, converting to axis/angle involves some nasty square roots and acos and so it's not exactly fast. I saw Eigen for the first time today, so I have no idea in what representation the Isometry3d type stores the data :D

I did like this because the `AbstractBasicTranslationRotation3D` of Magnum does not allow setting the transformation matrix. `Eigen::Isometry3d` is just a matrix that allows storing isometric transformations (i.e., rigid body). How can I set the matrix of `AbstractBasicTranslationRotation3D`? Another thing would be to check if Dart has some helpers to get it in Quaternions or something..

src/Magnum/DartIntegration/ConvertShapeNode.cpp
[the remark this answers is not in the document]
mosra: Wait, so it can happen that `getType()` returns `"BoxShape"`, but the type is *not* `BoxShape`? I hope not! :fearful:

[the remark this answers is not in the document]
costashatz: It should never happen according to Dart docs. So I will not check! :)

src/Magnum/DartIntegration/ConvertShapeNode.h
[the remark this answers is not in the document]
mosra: Because the users usually know better if given pointer is valid or not, so they can in many cases omit the check altogether. 

I see it as either needing to check for `nullptr` *before* or handling the return value being `NullOpt` *after* (and then maybe wondering what was the cause). In this case I think the former is simpler and thus better.

src/Magnum/DartIntegration/DartSkeleton.h
mosra: Wouldn't one usually use it in a loop? Then it doesn't matter, as implicit conversion makes everything sane again:

```cpp
for(DartObject& o: skel.bodyObjects()) {
    ...
}
```

(To explain myself: I'm trying hard to avoid using pointers for references that can never be null. From my experience, it always led to much simpler error handling and less uncertainty when interacting with such APIs -- "great, this can never be null" vs. "this is a pointer, i need to check the docs if it can be null")

src/Magnum/DartIntegration/ConvertShapeNode.cpp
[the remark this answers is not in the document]
mosra: So then you can also use `std::static_pointer_cast`, as dynamic cast is only a type check on top of it :)

src/Magnum/DartIntegration/ConvertShapeNode.h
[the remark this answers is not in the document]
costashatz: > I see it as either needing to check for nullptr before or handling the return value being NullOpt after (and then maybe wondering what was the cause). In this case I think the former is simpler and thus better.

Sounds good! It's also better for everything Magnum related to have similar styles.. :smile:

src/Magnum/DartIntegration/DartSkeleton.h
costashatz: > To explain myself: I'm trying hard to avoid using pointers for references that can never be null. From my experience, it always led to much simpler error handling and less uncertainty when interacting with such APIs -- "great, this can never be null" vs. "this is a pointer, i need to check the docs if it can be null"

As I said above, I have no reason to disagree with you and it's better for everything Magnum related to have the same logic/style.

src/Magnum/DartIntegration/ConvertShapeNode.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    // Get material information -- we ignore the alpha value
+    // Note that if this is a MeshShape we will ignore this material
+    Eigen::Vector4d col = shapeNode->getVisualAspect()->getRGBA();
+
+    // default shininess to 80.f
+    data._matData = new Trade::PhongMaterialData{Trade::PhongMaterialData::Flags{}, 80.f};
+    data._matData->ambientColor() = Color3{Float(col(0)), Float(col(1)), Float(col(2))};
+    data._matData->diffuseColor() = data._matData->ambientColor();
+    data._matData->specularColor() = data._matData->ambientColor();
```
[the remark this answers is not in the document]
costashatz: > Default for ambient is fully black, default for specular is fully white.

If I do not set them, I get random ambient/specular colors! But of course, I can set them to default values that make sense ;).

src/Magnum/DartIntegration/ConvertShapeNode.h
```diff
@@ -0,0 +1,82 @@
...
+*/
+struct VisualData {
+    /** mesh information */
+    Trade::MeshData3D* _meshData;
+    /** material information */
+    Trade::PhongMaterialData* _matData;
+    /** ImageData2D (vector) */
+    std::vector<Trade::ImageData<2>*> _imgData;
+    /** TextureData (vector) */
+    std::vector<Trade::TextureData*> _texData;
```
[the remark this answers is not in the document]
costashatz: > All types used here are movable (though not copyable) and, based on our previous discussion about uniqueness of the data, I think the structure could just contain everything directly as a value, without a pointer, even the vector -- nothing's ever copied, it's just moved around.
> 
> Feel free to correct me if I'm wrong :)

The only issue that I see is the following (you can help me by-pass it):

```cpp
for (UnsignedInt i = 0; i < importer->textureCount(); ++i) {
    Containers::Optional<Trade::TextureData> textureData = importer->texture(i);
    if (!textureData || textureData->type() != Trade::TextureData::Type::Texture2D) {
        Warning{} << "Cannot load texture, skipping";
        /* we need this to keep the indexing correct */
        data._imgData.emplace_back(nullptr);
        data._texData.emplace_back(nullptr);
        continue;
    }

    Containers::Optional<Trade::ImageData2D> imageData = importer->image2D(textureData->image());
    if (!imageData) {
        Warning{} << "Cannot load texture image, skipping";
        /* we need this to keep the indexing correct */
        data._imgData.emplace_back(nullptr);
        data._texData.emplace_back(nullptr);
        continue;
    }

    data._imgData.emplace_back(new Trade::ImageData2D{std::move(*imageData)});
    data._texData.emplace_back(new Trade::TextureData{std::move(*textureData)});
}
```

In this part of the code, I am reading the textures from the Mesh (`aiScene`). Just before I read the materials. If a material has textures, you can find them using an index. This index needs to be reserved (i.e., even if a texture could not be loaded, we would like all the others to be properly found and indexed). For this reason, I add this `nullptr` in cases where some texture could not be loaded so that the user can make the check and use all the other textures properly.

Another way of doing this, would be to directly alter the indexes of the materials (when needed). Do you have anything else to suggest?

src/Magnum/DartIntegration/ConvertShapeNode.cpp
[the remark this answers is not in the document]
mosra: That's a bug, then -- there should never be random data unless you explicitly ask for it. Sorry. Will fix this.

src/Magnum/DartIntegration/ConvertShapeNode.h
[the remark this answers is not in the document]
mosra: Right, sorry, missed this case for some reason.

What about having the structure like this, then?

```cpp
struct VisualData {
    /** mesh information */
    Trade::MeshData3D _meshData;

    /** material information */
    Trade::PhongMaterialData _matData;

    /** ImageData2D (vector) */
    std::vector<Containers::Optional<Trade::ImageData2D>> _imgData;

    /** TextureData (vector) */
    std::vector<Containers::Optional<Trade::TextureData>> _texData;
};
```

That would preserve the "nullability" of particular elements, but also avoids those heap allocations. I thought about reindexing as well, but came to a conclusion that it's better to preserve the "gaps" to indicate that not everything was correctly imported.

costashatz:
@mosra I think I addressed all the issues you raised! Thanks for the extensive review! Let me know if you have any more comments..!

Two things that remain are:

- Handling DART versions
- CI integration

I will try to address the first one as soon as possible and then we can see the second one!

mosra:
Thank you for the fast responses! Will get back to it this evening, hopefully :)

mosra:
Okay, I am very happy with the code now :) Last things that came to my mind:

- Does DART provide a `FindDART.cmake` module? Because there's a [builtin FindDart.cmake](https://cmake.org/cmake/help/v3.9/module/FindDart.html) and I doubt it's for DART simulation :) Because of the minor casing difference (that on windows would cause cmake to use the builtin module and then fail horribly), I think we should bundle `FindDART.cmake` in the `modules/`  directory in order to avoid such errors when DART is not installed. But I see only [CMake configs](https://github.com/dartsim/dart/tree/master/cmake) in their repo and no Find module. (I *hate* this aspect of CMake, argh.) I don't know how well you are with CMake, so unless you want to tackle this yourself I can look into it and try to provide some proxy `FindDART,cmake` that just redirects to their config module.
- CI: if you are willing to build DART for Travis (since the PPA with DART packages is not whitelisted and [since whitelisting new PPAs is not possible anymore since almost a year ago](https://github.com/travis-ci/apt-source-whitelist/blob/master/README.md), argh, Travis!!), that would be great. If you don't want to, I fully understand, it's a messy job ;)
- One last API breaking change -- sorry for not coming up with that earlier -- I think it would be better to name the `VisualData` members like this:

    ```cpp
    struct VisualData {
        /** mesh information */
        Trade::MeshData3D mesh;
        /** material information */
        Trade::PhongMaterialData material;
        /** ImageData2D (vector) */
        Containers::Array<Containers::Optional<Trade::ImageData2D>> images;
        /** TextureData (vector) */
        Containers::Array<Containers::Optional<Trade::TextureData>> textures;
    };
    ```

    I.e., not using the underscore (these are public members) and naming them without the `*Data` suffix, as "Data" is already contained in the struct name. Since this directly affects the example (and probably your other code as well), I thought it's better if you do this instead of me breaking the API later.

costashatz:
> Does DART provide a FindDART.cmake module? Because there's a builtin FindDart.cmake and I doubt it's for DART simulation :) Because of the minor casing difference (that on windows would cause cmake to use the builtin module and then fail horribly), I think we should bundle FindDART.cmake in the modules/ directory in order to avoid such errors when DART is not installed. But I see only CMake configs in their repo and no Find module. (I hate this aspect of CMake, argh.) I don't know how well you are with CMake, so unless you want to tackle this yourself I can look into it and try to provide some proxy FindDART,cmake that just redirects to their config module.

I am not as good as you, but I can give it a try (I am not horribly bad at CMake). I have to do the versioning thing also. If you have any hints of what I should do (for redirecting to their config module), please shoot. Overall, I certainly dislike CMake and I prefer python-based build systems ([waf](https://waf.io/), [scons](http://scons.org/), etc.), but CMake won the market so we have to bear with it! :stuck_out_tongue: 

> CI: if you are willing to build DART for Travis (since the PPA with DART packages is not whitelisted and since whitelisting new PPAs is not possible anymore since almost a year ago, argh, Travis!!), that would be great. If you don't want to, I fully understand, it's a messy job ;)

I will give this a try.

> I.e., not using the underscore (these are public members) and naming them without the *Data suffix, as "Data" is already contained in the struct name

That's true. You're right. Now that we are on it, maybe we can rename the whole structure as well (`VisualData` is not the ideal name for it). I was thinking `DartShapeData` or something like this (maybe just `ShapeData` as it is inside the `DartIntegration` namespace). What do you think?

mosra:
> If you have any hints of what I should do

I have no idea, actually :D I'm doing the inverse (including the Find module from the config module) in [MagnumConfig.cmake](https://github.com/mosra/magnum/blob/master/modules/MagnumConfig.cmake) (I don't even remember anymore why exactly this was needed and what CMake "speciality" was this working around again, but whatever) so a similar thing could be done in the `FindDART.cmake` proxy and could "just work", hopefully. But there you don't know the location of their config module at all -- would need to search for it first probably (and then produce some nice error message if not found).

> I will give this a try.

Thank you :+1: 

> `ShapeData`

Yup, `ShapeData` is nice :+1:

costashatz:
> Does DART provide a FindDART.cmake module

I was examining this and I realized that when installing DART, you get the following files:

```bash
/usr/local/lib/pkgconfig/dart.pc 

/usr/local/share/dart/cmake/DARTConfig.cmake
/usr/local/share/dart/cmake/DARTConfigVersion.cmake
```

The `dart.pc` is the one used to point to the other files. I have a feeling that we should not do anything and they (the developers of DART) took care of it.

mosra:
I wanted to avoid this problem:

```
CMake Error at CMakeLists.txt:55 (find_package):
  By not providing "FindDART.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "DART", but
  CMake did not find one.

  Could not find a package configuration file provided by "DART" with any of
  the following names:

    DARTConfig.cmake
    dart-config.cmake

  Add the installation prefix of "DART" to CMAKE_PREFIX_PATH or set
  "DART_DIR" to a directory containing one of the above files.  If "DART"
  provides a separate development package or SDK, be sure it has been
  installed.
```

This is what the people get when they don't have DART installed. Incredibly confusing message that CMake for some reason spits out instead of just saying "sorry, something called DART is not installed". This gets even worse on Windows, where due to the case insensivity of the file system an error similar to the following will be produced:

```
CMake Error at /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find Dart (missing: DART_ROOT)
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.10/Modules/FindDart.cmake:30 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:55 (find_package)
```

But -- let's keep this as-is for now. The solution doesn't seem very obvious at the moment and it can be always fixed later when people actually start complaining ;ú

costashatz:
@mosra can you check if what I did makes sense and if it works (on windows it should not!)? Thanks!

mosra:
This will inevitably get complicated and I don't think we're able to provide a solution that's flexible enough ... :/

Got another idea (way too late, sorry): it's possible to say `find_package(DART CONFIG REQUIRED)`, which will produce a bit shorter error message in case DART is not found and also behaves correctly on Windows (not aliasing with Dart anymore). I think we could live with that. So, instead of providing `FindDART.cmake`, could you just add `CONFIG` to all `find_package(DART)` commands?

costashatz:
> Got another idea (way too late, sorry): it's possible to say find_package(DART CONFIG REQUIRED), which will produce a bit shorter error message in case DART is not found and also behaves correctly on Windows (not aliasing with Dart anymore). I think we could live with that. So, instead of providing FindDART.cmake, could you just add CONFIG to all find_package(DART) commands?

Yes..! Why not?

mosra:
Merged into `master` now. Thank you for the contribution, looking forward to the example ;)

PR #29, opened 2018-02-09
[WIP]: Dart integration v2

costashatz:
This PR improves the DartIntegration introduced in #28 .. It is still work in progress, but at a quite good state:

- [x] Require DART version >=6.0.0; I am highly unsure if the code works with older versions of DART
- [x] Check for mesh updates from DART
- [x] Support soft bodies
- [x] Incorporate all the DART flags for rendering
- [ ] Support multiple meshes and materials per `DartIntegration::Object`
- [ ] Provide default behavior (i.e, remove it from scene) when an Object is deleted from the DART world; maybe we do not need to add this
- [ ] Fix normals in DART `SoftMeshShape`; for now we are rendering them without normals
- [ ] Code optimization
    - [ ] Properly edit already created meshes (e.g., do not re-create them from scratch when just scaling needs to be changed)
    - [ ] Check the usage of `std::shared_ptr`; maybe replace it with `std::unique_ptr`? It could not be a good idea to have `unique_ptr`s as you may need to use the pointers to check whether a mesh is already created/or not

For an example usage, you can have a look at [this](https://github.com/costashatz/magnum-examples/tree/dart-example) branch, which I am updating to match the updates on this integration.

Let me know what you think!


src/Magnum/DartIntegration/Object.h
```diff
@@ -40,72 +45,93 @@ namespace dart { namespace dynamics {
...
+Object* obj = new Object{&object, body};
 @endcode
 
 or
 
 @code{.cpp}
 dart::dynamics::ShapeNode* node = getShapeNodeFromDart();
 SceneGraph::Object<SceneGraph::MatrixTransformation3D> object;
-DartObject* obj = new DartObject{&object, node};
+Object* obj = new Object{&object, node};
```
mosra: After this rename I would prefer to use the explicit namespace in the documentation snippets (`DartIntegration::Object`), similarly to what's done with SceneGraph APIs.

src/Magnum/DartIntegration/Object.h
```diff
@@ -40,72 +45,93 @@ namespace dart { namespace dynamics {
     class ShapeNode;
 }}
 
-namespace Magnum { namespace DartIntegration {
+namespace Magnum {
+
+class Mesh;
+class Buffer;
+template<UnsignedInt dimensions> class Texture;
```
mosra: Please use one of the [forward declaration headers](http://doc.magnum.graphics/magnum/compilation-speedup.html#compilation-forward-declarations) instead of specifying the forward declarations by hand. It gets very messy very fast once the templates get involved :)

In this case it would be `#include <Magnum/Magnum.h>`.

src/Magnum/DartIntegration/Object.h
```diff
@@ -40,72 +45,93 @@ namespace dart { namespace dynamics {
...
+
+    /** @brief index Buffer */
+    Buffer* indexBuffer;
+
+    /** @brief Material data */
+    /* @todo: support multiple materials */
+    Trade::PhongMaterialData material;
+
+    /** @brief Textures */
+    Containers::Array<Texture<2>*> textures;
```
mosra: Continuing with the forward declaration header from above, you could then simply use `Texture2D` here.

src/Magnum/DartIntegration/Object.h
```diff
@@ -114,11 +140,36 @@ class MAGNUM_DARTINTEGRATION_EXPORT DartObject: public SceneGraph::AbstractBasic
...
+        -   `SphereShape`
+
+        The following DART shapes are not yet supported:
+
+        -   `ConeShape`
+        -   `LineSegmentShape`
+        -   `MultiSphereConvexHullShape`
+        -   `PlaneShape` (this is an infinite plane with normal)
+        */
+        bool _convertShapeNode();
```
mosra: This is a `private` function and so the documentation will get completely ignored. Can you put that to the class docs instead?

Also, please don't prefix *functions* with underscore. The func is private already, that's enough :)

src/Magnum/DartIntegration/Test/CMakeLists.txt
```diff
@@ -42,9 +43,9 @@ endif()
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/configure.h.cmake
                ${CMAKE_CURRENT_BINARY_DIR}/configure.h)
 
-corrade_add_test(DartIntegrationDartSkeletonTest DartSkeletonTest.cpp LIBRARIES MagnumDartIntegration ${DART_LIBRARIES})
-target_include_directories(DartIntegrationDartSkeletonTest PUBLIC ${DART_INCLUDE_DIRS})
+corrade_add_test(DartIntegrationDartWorldTest DartWorldTest.cpp LIBRARIES Magnum::OpenGLTester MagnumDartIntegration ${DART_LIBRARIES})
```
mosra: Because the test needs GL now, in order to keep things running properly on particular systems and the CIs (which don't have access to GL at the moment) there needs to be:

- a new `BUILD_GL_TESTS` CMake option (see [here](https://github.com/mosra/magnum-extras/blob/12d2310e177019ccb6b1f40cacd95817c9856df2/CMakeLists.txt#L62) for how's that done elsewhere -- search for BUILD_GL_TESTS in that file)
- this test named `DartIntegrationWorldGLTest` (so it can be easily black/whitelisted during `ctest` run)
- this test added only `if(BUILD_GL_TESTS)` (see e.g. [here](https://github.com/mosra/magnum/blob/52613a2ad94e7c6f11c7898e162f636db7d07980/src/Magnum/Test/CMakeLists.txt#L130) for an example), don't forget to split out the `set_target_properties()` command below as well

Sorry that most of the supporting structures for GL tests are not here yet -- it was just not needed (or not cared for) until now :)

src/Magnum/DartIntegration/Test/CMakeLists.txt
```diff
@@ -42,9 +43,9 @@ endif()
...
+target_include_directories(DartIntegrationDartWorldTest PUBLIC ${DART_INCLUDE_DIRS})
```
mosra: Now, thinking about this, isn't adding DART include dirs done implicitly through the transitive dependency on DART?

src/Magnum/DartIntegration/Object.h
costashatz: > After this rename I would prefer to use the explicit namespace in the documentation snippets (DartIntegration::Object), similarly to what's done with SceneGraph APIs.

Good point!

src/Magnum/DartIntegration/Object.h
costashatz: > Please use one of the forward declaration headers instead of specifying the forward declarations by hand. It gets very messy very fast once the templates get involved :)
> 
> In this case it would be #include <Magnum/Magnum.h>.

I should at some point read in more detail the docs! :smile:

src/Magnum/DartIntegration/Object.h
costashatz: > This is a private function and so the documentation will get completely ignored. Can you put that to the class docs instead?

Of course...

src/Magnum/DartIntegration/Test/CMakeLists.txt
costashatz: > Now, thinking about this, isn't adding DART include dirs done implicitly through the transitive dependency on DART?

Good catch! Thanks!

src/Magnum/DartIntegration/Test/CMakeLists.txt
costashatz: > Sorry that most of the support code for GL tests is not here yet -- it was not needed (or not cared for) until now :)

No problem! Should I also do the `-E GLTest` as [here](https://github.com/mosra/magnum-extras/blob/master/package/ci/travis-desktop.sh#L55)?

mosra: Yes, please. I thought at least that is done already, but apparently just partially -- in particular [this line](https://github.com/mosra/magnum-integration/blob/bb743603b1910da2936c387fd59bcabfb5d67b70/package/ci/travis-desktop.sh#L74) should say `-R Dart -E GLTest`.

costashatz:
@mosra I think it's time that you can check more thoroughly what I've written and give me your thoughts/comments on the code and the 2 last bullets remaining (about the pointers).. Of course I will work on the documentation more.. :smile: 

Thanks!

src/Magnum/DartIntegration/Object.h
```diff
@@ -0,0 +1,183 @@
...
+        /** @brief Underlying DART `ShapeNode` */
+        dart::dynamics::ShapeNode* shapeNode() { return _node; }
+
+        /** @brief Underlying DART `BodyNode` */
+        dart::dynamics::BodyNode* bodyNode() { return _body; }
+
+    private:
+        explicit Object(SceneGraph::AbstractBasicObject3D<Float>& object, SceneGraph::AbstractBasicTranslationRotation3D<Float>& transformation, dart::dynamics::ShapeNode* node, dart::dynamics::BodyNode* body);
+
+        bool convertShapeNode();
```
mosra: So, this comment is why the review took so long (sorry about that): I tried to come up with a solution to make this whole thing more testable because otherwise it's not really possible to verify that everything is working correctly -- all you get now are some opaque Buffer and Mesh structures. I think I have it now. So, what about this: 

Previously, there was a `convertShapeNode()` function that didn't depend on GL context being present and which just returned the populated `ShapeData`. This function was very testable, so I'd vote for restoring it (together with the header it was), but with an updated API:

- it takes the `ShapeNode&`, as previously, but has no overload for `DartIntegration::Object`, as this class handles it internally
- it takes an [EnumSet](http://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1EnumSet.html) to describe which data should be extracted and which not so the Object can update only what's needed
- it takes an existing instance of plugin manager instead of creating an instance of it inside

The Object would call `convertShapeNode()` inside, asking for extraction of only the data that changed, and then it calls `MeshTools::compile()` to populate its own internal mesh, buffer etc. instances. (So, yes, the `ShapeData` would contain just the data again.)

Instead of having the plugin manager as `static` variable inside a function, the instance would be part of the World instance and every Object would have access to it. (Reason: `static` variables inside functions are *very evil* -- they cause the compiler to generate thread-safe wrapper around these even if you aren't involve with any threading, the destruction of the whole variable is at some unspecified point after exiting from `main()`, etc. etc.)

This way the majority of the code could be properly tested again (maxing out the code coverage, verifying how the imported mesh data actually look etc.). Besides that, this would make this library less tightly integrated with SceneGraph, allowing users to implement the transformation handling in their own way, if they want to.

I hope I didn't miss anything that would make this impossible. As always, I'm open for further discussion :)

src/Magnum/DartIntegration/Test/WorldGLTest.cpp
```diff
@@ -0,0 +1,514 @@
...
+    explicit DartIntegrationTest();
+
+    void test();
+    void test_soft();
+    void urdf();
+    void test_multi_mesh();
+    void texture();
+
+    void bench_simple();
+    void bench_soft();
```
mosra: Coding style: use `camelCase` instead of `underscore_case`, please. Also no need to prefix with `test` or `bench`, it's clear enough from the test output :)

src/Magnum/DartIntegration/Test/CMakeLists.txt
```diff
@@ -42,9 +43,8 @@ endif()
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/configure.h.cmake
                ${CMAKE_CURRENT_BINARY_DIR}/configure.h)
 
-corrade_add_test(DartIntegrationDartSkeletonTest DartSkeletonTest.cpp LIBRARIES MagnumDartIntegration ${DART_LIBRARIES})
-target_include_directories(DartIntegrationDartSkeletonTest PUBLIC ${DART_INCLUDE_DIRS})
+corrade_add_test(DartIntegrationWorldGLTest WorldGLTest.cpp LIBRARIES Magnum::OpenGLTester MagnumDartIntegration ${DART_LIBRARIES})
```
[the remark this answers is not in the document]
costashatz: Here we need to link to DART libs, because for the test we require more (i.e., either `utils-urdf` or `io-urdf`) libs from DART than the rest of the integration and as such cannot link properly..

src/Magnum/DartIntegration/Object.cpp
```diff
@@ -0,0 +1,519 @@
...
+  for(std::size_t i=0; i<bn->getNumFaces(); ++i)
+  {
+    const Eigen::Vector3i& face = bn->getFace(i);
+    for(std::size_t j=0; j<3; ++j)
+      normals[face[j]] += normalFromVertex(bn, face, j);
+  }
+
+  for(std::size_t i=0; i<normals.size(); ++i)
+    normals[i].normalize();
+}
```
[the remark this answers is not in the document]
costashatz: > This whole thing could be replaced by MeshTools::generateFlatNormals(), I think. Correct me if I'm wrong, of course ;)

Initially I was using the `MeshTools::generateFlatNormals()` functionality, but then I realized that their SoftMeshes are ''special''. Long story short, the triangles generated in the mesh do not all follow the same orientation (i.e., either clockwise or anti-clockwise) and as such I found this code in their own GUI implementations in order to try to fix this issue. The result is still not the best:

image

Nevertheless, it is much better than generating the normals assuming that they follow some orientation; in that case, we get wrong lighting (e.g., triangles completely black etc)..

As a side note, I found this [nice library](https://github.com/libigl/libigl) that has a nice [facet orientation fixing algorithm](http://libigl.github.io/libigl/tutorial/tutorial.html#facetorientation). I might  at some point re-implement their algorithm for Magnum (it might be a nice feature) and thus we could use it here to get much nicer normals..

src/Magnum/DartIntegration/Object.cpp
```diff
@@ -0,0 +1,519 @@
...
+Object& Object::update() {
+    if(_node)
+        convertShapeNode();
+    _used = true;
+    /* Get transform from DART */
+    const Eigen::Isometry3d* trans;
+    if(!_node)
+        trans = &_body->getRelativeTransform();
+    else
+        trans = &_node->getRelativeTransform();
```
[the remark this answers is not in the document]
mosra: I really need to dig deeper into Eigen :D I tried to understand what Isometry3d actually is a few times, but always failed :)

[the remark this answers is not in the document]
costashatz: > I tried to understand what Isometry3d actually is a few times, but always failed :)

`Eigen::Isometry3d` is an alias for [Eigen::Transform](https://eigen.tuxfamily.org/dox/classEigen_1_1Transform.html) that allows only [isometric](https://en.wikipedia.org/wiki/Isometry) transformations.

src/Magnum/DartIntegration/Object.h
costashatz: > Previously, there was a convertShapeNode() function that didn't depend on GL context being present and which just returned the populated ShapeData. This function was very testable, so I'd vote for restoring it (together with the header it was), but with an updated API:
> 
> it takes the ShapeNode&, as previously, but has no overload for DartIntegration::Object, as this class handles it internally
> it takes an EnumSet to describe which data should be extracted and which not so the Object can update only what's needed
> it takes an existing instance of plugin manager instead of creating an instance of it inside
> The Object would call convertShapeNode() inside, asking for extraction of only the data that changed, and then it calls MeshTools::compile() to populate its own internal mesh, buffer etc. instances. (So, yes, the ShapeData would contain just the data again.)

Thanks for the comments! I am working on this and I should have something by tomorrow. I do not see any reason for now why this would not work...

src/Magnum/DartIntegration/Object.cpp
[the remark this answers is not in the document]
mosra: I mean, like the internal representation, if it's a quat + translation, or a matrix or what exactly. But, nevermind :)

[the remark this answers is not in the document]
costashatz: Never really wondered about it, but it [seems that it is a matrix](https://bitbucket.org/eigen/eigen/src/de7544f256bdeb135f7d016e2ddf344a9e0406eb/Eigen/src/Geometry/Transform.h?at=default&fileviewer=file-view-default#Transform.h-250) :D..

src/Magnum/DartIntegration/Object.cpp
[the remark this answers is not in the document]
mosra: Right, sorry, I remember we were talking about this.

I looked at the algorithm only briefly at first and thought it's generating flat normals (didn't see the `+=` at the end). Now I see that it's generating smooth normals: averaging the normal over neighboring faces, doing *some* weighting on top (I'm not sure if completely correctly, as I see some weird artifacts on the screenshot). That fixes the face winding issue only by accident, I think -- only if there's not many faces with wrong winding.

Compared to the flat normals the result is quite ugly, though :/ 

Most importantly, why I'm commenting: this code snippet is never going to be tested well enough in the context of the DartIntegration library and since it doesn't really solve the problem in a nice way, I don't know if it's worth to invest further time in this direction (I originally wanted to suggest extracting this code out into MeshTools and test it there).

But I got two different ideas:

1. what if you would, in case of those "special SmoothMeshes", add every face twice, once in the original winding and once in the reverse? That, together with enabling face culling and generating flat normals afterwards, could solve the issue in a nice-looking (yet ugly) way. This would need to be explicitly mentioned in the docs, as forgetting to enable face culling may have disastrous results, heh :)
2. what if you would use functionality from Assimp itself to generate the normals? That would remove the need for the above code, avoid extra testing effort and we could blame a third-party lib for producing ugly artefacts :laughing: The [relevant option is here](http://assimp.sourceforge.net/lib_html/postprocess_8h.html#a64795260b95f5a4b3f3dc1be4f52e410a6afb4fee42eca4482674859196cb8685). But I give no guarantees that this approach would also *accidentally circumvent* the face winding issue (and it will still look ugly). So maybe solution 1 is still the best among the worst :D

By the way, did you report this "speciality" to DART? Fixing it there would be the best solution in my opinion.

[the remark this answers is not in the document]
mosra: I did a bit of research, opened https://github.com/mosra/magnum/pull/229 so I don't forget all the details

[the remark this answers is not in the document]
costashatz: > what if you would, in case of those "special SmoothMeshes", add every face twice, once in the original winding and once in the reverse? That, together with enabling face culling and generating flat normals afterwards, could solve the issue in a nice-looking (yet ugly) way. This would need to be explicitly mentioned in the docs, as forgetting to enable face culling may have disastrous results, heh :)

I think I like this idea more..! Let me finish the `convertShapeNode()` first and then I'll do this.. Thanks for the suggestions..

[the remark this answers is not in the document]
costashatz: > By the way, did you report this "speciality" to DART? Fixing it there would be the best solution in my opinion.

They already know it! :stuck_out_tongue: I can make a more explicit issue with possible solutions (or even fix it).. Putting it on my TO-DO list..!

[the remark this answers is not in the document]
mosra: I tried to search, but didn't find any issue related to this on their main repo.

Adding a link to the issue to the documentation about "why is everything twice and why one needs to enable face culling" would be great, so somebody can look it up later and discard the whole workaround once they fix it :)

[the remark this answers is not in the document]
costashatz: > Adding a link to the issue to the documentation about "why is everything twice and why one needs to enable face culling" would be great, so somebody can look it up later and discard the whole workaround once they fix it :)

The thing is that they do not really care about the graphics part of the library. They have it (and update it) only when they need something (or someone does it for them). That's why I said it's better if I implement it myself for them.. :stuck_out_tongue:

src/Magnum/DartIntegration/World.h
```diff
@@ -0,0 +1,240 @@
...
+            dartObjectCreator = [](SceneGraph::AbstractBasicObject3D<Float>& parent, dart::dynamics::BodyNode* body) -> std::shared_ptr<Object> {
+                return std::make_shared<Object>(static_cast<T&>(parent), body);
+            };
+            dartShapeObjectCreator = [](SceneGraph::AbstractBasicObject3D<Float>& parent, dart::dynamics::ShapeNode* node) -> std::shared_ptr<Object> {
+                return std::make_shared<Object>(static_cast<T&>(parent), node);
+            };
+
+            refresh();
+        }
+
```
mosra: The class has some heavy members, I would add an explicit destructor here and `= default` it in the `*.cpp` file to speed up the compilation.

Similarly for the constructor -- delegate most of the initialization to a private non-templated constructor that has a definition in the `*.cpp` file.

src/Magnum/DartIntegration/ConvertShapeNode.cpp
```diff
@@ -26,167 +26,293 @@
...
+#include <Magnum/Trade/MeshData3D.h>
+#include <Magnum/Trade/MeshObjectData3D.h>
+#include <Magnum/Trade/ObjectData3D.h>
+#include <Magnum/Trade/PhongMaterialData.h>
+#include <Magnum/Trade/TextureData.h>
 
 namespace Magnum { namespace DartIntegration {
 
-Containers::Optional<ShapeData> convertShapeNode(dart::dynamics::ShapeNode& shapeNode) {
+Containers::Optional<ShapeData> convertShapeNode(dart::dynamics::ShapeNode& shapeNode, ShapeLoadTypes loadType, Trade::AbstractImporter* importer) {
```
mosra: The `importer` could be just a reference, not a pointer. Similarly in all functions that call `convertShapeNode()` -- `Object::update()` etc.

src/Magnum/DartIntegration/ConvertShapeNode.cpp
```diff
@@ -26,167 +26,306 @@
...
+                /* fallback material to by-pass seg-faults */
+                else
+                    new(&materials[j]) Trade::PhongMaterialData{Trade::PhongMaterialData::Flags{}, 80.f};
+
+                if(getMesh) {
+                    new(&meshes[j]) Trade::MeshData3D{std::move(*meshData)};
+                }
+                /* fallback mesh data to by-pass seg-faults */
+                else
+                    new(&meshes[j]) Trade::MeshData3D{MeshPrimitive::Triangles, std::vector<UnsignedInt>(), std::vector<std::vector<Vector3>>(1, std::vector<Vector3>()), std::vector<std::vector<Vector3>>(), std::vector<std::vector<Vector2>>(), std::vector<std::vector<Color4>>()};
```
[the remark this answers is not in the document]
costashatz: > One more argument for Optional: if someone is using the ShapeData directly, they currently wouldn't have a intuitive way to know if a particular mesh is valid or not. With Optional this would be obvious.

If a mesh is asked to be loaded and cannot be loaded, we return `Containers::NullOpt`. Otherwise the size of the `ShapeData::meshes` gives us the number of loaded meshes (`Trade::MeshData3D`) and all of them are always valid..

> As we discussed before -- wouldn't Containers::Array<Containers::Optional<Trade::MeshData3D> (and similar for materials) be a better / safer solution here, to avoid such suspiciously-looking workarounds?

To this I mostly agree, so I am also inclining to change things to `Containers::Optional`..

src/Magnum/DartIntegration/Object.cpp
```diff
@@ -0,0 +1,209 @@
...
+    bool firstTime = !_drawData;
+
+    dart::dynamics::ShapeNode& shapeNode = *this->shapeNode();
+    dart::dynamics::ShapePtr shape = shapeNode.getShape();
+
+    bool getMaterial = shape->checkDataVariance(dart::dynamics::Shape::DataVariance::DYNAMIC_COLOR);
+    bool getPrimitive = shape->checkDataVariance(dart::dynamics::Shape::DataVariance::DYNAMIC_PRIMITIVE);
+    bool getMesh = shape->checkDataVariance(dart::dynamics::Shape::DataVariance::DYNAMIC_VERTICES)
+                    || shape->checkDataVariance(dart::dynamics::Shape::DataVariance::DYNAMIC_ELEMENTS)
+                    || shape->checkDataVariance(dart::dynamics::Shape::DataVariance::DYNAMIC);
```
mosra: Almost there: these three bools could be moved directly inside the `if(get...)` conditions below, as they are not used anywhere else

src/Magnum/DartIntegration/Object.cpp
```diff
@@ -0,0 +1,209 @@
...
+            new(&_drawData->meshes[i]) Mesh{std::move(mesh)};
+            _drawData->vertexBuffers[i] = std::move(*vertexBuffer.release());
+            if(indexBuffer)
+                _drawData->indexBuffers[i] = std::move(*indexBuffer.release());
+        }
+    }
+
+    /* If we got here, everything went OK;
+     * update flag for rendering */
+    _updatedMesh = firstTime || getMaterial || getPrimitive || getMesh;
```
mosra: this could be simply `_updatedMesh = !!loadType`

src/Magnum/DartIntegration/World.cpp
```diff
@@ -0,0 +1,170 @@
...
+    for(auto& obj: _dartToMagnum)
+        if(obj.second->bodyNode())
+            objs.emplace_back(std::ref(*obj.second));
+
+    return objs;
+}
+
+std::reference_wrapper<Object> World::objectFromDartFrame(dart::dynamics::Frame* frame) {
+    return std::ref(*_dartToMagnum[frame]);
+}
```
mosra: Sorry to sound like a broken record here: you can return just `Object&` here, no need for the wrapper ;)

Also: since this is a public API at the moment, think about what happens when you ask for a `frame` that's not in the `_dartToMagnum` structure: it will insert a new element with such key, default-construct it (so you get `nullptr` value) and then you dereference it ... :boom: 

Suggestion:

- I would use `at()` instead of `operator[]`, as that does not insert (but rather throw) when not found
- I would explicitly mention the behavior in the docs

src/Magnum/DartIntegration/ConvertShapeNode.cpp
```diff
@@ -26,167 +26,319 @@
...
 
-    if(shape->getType() == dart::dynamics::MeshShape::getStaticType()) {
-        /*  @todo check if we should not ignore the transformation in the Mesh */
+        if(loadType & ConvertShapeType::Mesh) {
+            shapeData.meshes = Containers::Array<Trade::MeshData3D>(Containers::NoInit, 1);
+            new(&shapeData.meshes[0]) Trade::MeshData3D{Primitives::Icosphere::solid(5)};
+        }
+    } else if(shape->getType() == dart::dynamics::MeshShape::getStaticType()) {
+        if (!importer) {
+            Error{} << Debug::boldColor(Debug::Color::Red) << "DartIntegration::convertShapeNode(): AssimpImporter is not available and you are trying to load a dart::dynamics::MeshShape" << Debug::resetColor;
```
mosra: Unconditional output coloring will make problems when redirecting output to a text file (or on CIs that don't understand color codes). So I would rather have the boring uncolored output here instead.

mosra:
Merged, thanks a lot. Code coverage including GL tests says 92%, you're doing amazing work :+1: :tada: 

I made a few small changes to the public API that might break your code, see e3efe6de0791242a53324873ce571003fc617712, 3c3d554f7330ddbbc52acdabff0ba33c35ed1f43 and 7676fdb34e22eb8edde46fdc3023edb524e408c5. In 016c2527f7da3b10a2277cf0be9e6b7cded5ad7e I removed a public function that I think was superfluous and never used in any code, I hope that's okay.

Looking forward to the example now :)

PR #31, opened 2018-05-27
fix to allow building on both Ubuntu 16.04 and 18.04.

matjam:
As promised in https://github.com/mosra/magnum/issues/248, here is a PR that will allow builds to work for both Ubuntu 18.04 and 16.04. Note that some of the changes are also to allow the build to work correctly in Launchpad, as it builds from a clean chroot every time dependencies may not be available.

### DIFF
--- a/package/debian/control
+++ b/package/debian/control
@@ -18,7 +18,7 @@ Description: Magnum Integration development files
 Package: magnum-integration
 Section: libs
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}, magnum, libbulletcollision2.83.6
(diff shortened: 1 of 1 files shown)


mosra:
Hmm. Unlike in the other PRs, here you don't mention `corrade` and `corrade-dev` among dependencies. Is that intentional?

matjam:
hmm, might not have been necessary?

I'd have to go and rebuild everything in launchpad to check ... yikes.

mosra:
Merged in 58d681bf9626aaa85815722524711ae591f7b5c4. Thanks! :)

PR #33, opened 2018-11-12
Move Imgui integration

Squareys:
Hi all!

This is the WIP PR for my efforts to move various magnum-imgui projects (by @lecopivo https://github.com/lecopivo/magnum-imgui and @denesik and forks by @nnarain, @williamjcm and @ShaddyDC) into the official repository.

Cheers, Jonathan

**TODOs:**
 - [x] Add code from other repos
 - [x] CMake setup
 - [x] Add to find module
 - [x] Migrate example => https://github.com/mosra/magnum-examples/pull/51
    - [x] Fix duplicate symbols issue with current find method
 - [x] Add a test
 - [x] Build on CI
 - [x] Adapt Code Style
 - [x] Add documentation
 - [x] Re-check mentions in the license headers


modules/FindMagnumIntegration.cmake
```diff
@@ -16,6 +16,7 @@
 #  Bullet                       - Bullet Physics integration library
 #  Dart                         - Dart Physics integration library
 #  Glm                          - GLM integration library
+#  ImGui                        - ImGui integration library
```
mosra: Having `doc/cmake-integration.dox` updated as well would be wonderful :)

CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ find_package(Magnum REQUIRED)
 option(WITH_BULLET "Build BulletIntegration library" OFF)
 option(WITH_DART "Build DartIntegration library" OFF)
 option(WITH_GLM "Build GlmIntegration library" OFF)
+option(WITH_IMGUI "Build ImGuiIntegration library" OFF)
```
mosra: Having `doc/building-integration.dox` updated as well would be wonderful :)

modules/FindMagnumIntegration.cmake
```diff
@@ -190,6 +191,14 @@ foreach(_component ${MagnumIntegration_FIND_COMPONENTS})
 
             set(_MAGNUMINTEGRATION_${_COMPONENT}_INCLUDE_PATH_NAMES MotionState.h)
 
+        # ImGui integration library
+        elseif(_component STREQUAL ImGui)
+            find_package(imgui)
```
mosra: where is the find module for this? is it distributed with imgui itself (as a config, maybe)?

mosra:
FYI, just found this: https://gitlab.com/caleb-vincent/MagnumImGui so maybe that has also some changes worth integrating (if you would open an issue there notifying the author about this merge, that would be great :+1:)

Squareys:
@mosra Issues are not open on that project. Found a change, though, that I will integrate :+1:

src/Magnum/ImGuiIntegration/Integration.h
```diff
@@ -42,31 +42,81 @@ namespace Magnum { namespace ImGuiIntegration {
 
 class MAGNUM_IMGUIINTEGRATION_EXPORT ImGuiShader : public GL::AbstractShaderProgram {
```
mosra: Does this even need to be exposed publicly? For Magnum::Ui I'm not really exposing the shaders either. 

 Then it wouldn't need any docs (and also any `EXPORT`, I think).

Squareys: I was thinking of moving it into an `Implementation` namespace even... would that make sense? Would more clearly convey that it's not API

mosra: Yes, definitely.

Squareys:
@mosra The code would be ready for some initial rough review pass. The documentation is not done yet, though.

modules/FindImGui.cmake
```diff
@@ -0,0 +1,93 @@
...
+#   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+#   THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#   DEALINGS IN THE SOFTWARE.
+#
+
+# vcpkg installs cmake package config files for imgui. We're making an INTERFACE
+# target for it, which is supported only since CMake 3.0.
+if(NOT CMAKE_VERSION VERSION_LESS 3.0.0)
```
mosra: Just assume CMake 3.0 is used, I'll be dropping 2.8.12 soon (right now in the process of removing GCC 4.7 support, this comes right after).

modules/FindMagnumIntegration.cmake
```diff
@@ -190,6 +191,14 @@ foreach(_component ${MagnumIntegration_FIND_COMPONENTS})
 
             set(_MAGNUMINTEGRATION_${_COMPONENT}_INCLUDE_PATH_NAMES MotionState.h)
 
+        # ImGui integration library
+        elseif(_component STREQUAL ImGui)
+            find_package(ImGui REQUIRED)
```
mosra: Not sure about the `REQUIRED` here -- if `find_package(MagnumIntegration OPTIONAL)` would be used, this will break. (But OTOH I also don't know what's the *proper* way of doing this, I'm only saying `find_package(Name)` everywhere.)

package/ci/appveyor.yml
```diff
@@ -65,6 +65,10 @@ install:
 - IF NOT EXIST %APPVEYOR_BUILD_FOLDER%\0.9.9.0.zip appveyor DownloadFile https://github.com/g-truc/glm/archive/0.9.9.0.zip
 - 7z x 0.9.9.0.zip -o%APPVEYOR_BUILD_FOLDER%\deps && ren %APPVEYOR_BUILD_FOLDER%\deps\glm-0.9.9.0 glm
 
+# ImGui
+- IF NOT EXIST %APPVEYOR_BUILD_FOLDER%\imgui-1.65.zip appveyor DownloadFile https://github.com/ocornut/imgui/archive/v1.65.zip
+- 7z x v1.65.zip -o%APPVEYOR_BUILD_FOLDER%\deps && ren %APPVEYOR_BUILD_FOLDER%\deps\imgui-1.65 imgui
```
mosra: Thinking out loud: we depend on particular version and since imgui backwards compatibility is broken quite often (at least by looking at changes in bgfx) it should be probably mentioned in the docs what version this integration supports.

src/Magnum/ImGuiIntegration/Integration.h
```diff
@@ -0,0 +1,197 @@
+#ifndef Magnum_ImGuiIntegration_Integration_h
+#define Magnum_ImGuiIntegration_Integration_h
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018
+              Vladimír Vondruš <mosra@centrum.cz>
+    Copyright © 2018 ShaddyAQN <ShaddyAQN@gmail.com>
+    Copyright © 2018 lecopivo <skrivantomas@seznam.cz>
```
mosra: `Copyright © 2018 Tomáš Skřivan <skrivantomas@seznam.cz>`  ;)

... and similarly elsewhere (if applicable)

src/Magnum/ImGuiIntegration/Integration.h
mosra: Also: unless you construct / destruct this class (or call any non-inline functions on it) in the public header (which looks like you don't), there's no need to `MAGNUM_IMGUIINTEGRATION_EXPORT` it.

src/Magnum/ImGuiIntegration/Integration.h
```diff
@@ -0,0 +1,197 @@
...
+@code{.cpp}
+Context context;
+
+// ...
+
+if(Context::get().newFrame({640, 480}, {640, 480})) {
+    // ...
+}
+
+@endcode
```
mosra: Would be great to have this snippet compiled -- put in a cpp file into `doc/snippets/`, mentioned in `doc/snippets/CMakeLists.txt` and referenced from here via `@snippet ImGuiIntegration.cpp something`. Look into BulletIntegration or OVR for reference.

src/Magnum/ImGuiIntegration/Integration.hpp
```diff
@@ -0,0 +1,187 @@
...
+#include <Magnum/GL/Extensions.h>
+
+#include <imgui.h>
+
+using namespace Magnum;
+
+namespace Magnum { namespace ImGuiIntegration {
+
+template<class KeyEvent>
+bool Context::handleKeyEvent(KeyEvent& event, bool value) {
```
mosra: no newlines after `template<>` 

and similarly below

src/Magnum/ImGuiIntegration/Integration.hpp
```diff
@@ -0,0 +1,187 @@
...
+#include <Magnum/PixelFormat.h>
+#include <Magnum/GL/Renderer.h>
+#include <Magnum/GL/Shader.h>
+#include <Magnum/GL/TextureFormat.h>
+#include <Magnum/GL/Version.h>
+#include <Magnum/GL/Extensions.h>
+
+#include <imgui.h>
+
+using namespace Magnum;
```
mosra: No need for this here ;)

src/Magnum/OvrIntegration/CMakeLists.txt
```diff
@@ -82,4 +82,4 @@ if(BUILD_TESTS)
 endif()
 
 # Magnum OVR integration target alias for superprojects
-add_library(MagnumIntegration::Ovr ALIAS MagnumOvrIntegration)
+
```
mosra: W...why this? :laughing:

src/Magnum/ImGuiIntegration/ImGui.vert
```diff
@@ -0,0 +1,61 @@
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018
+              Vladimír Vondruš <mosra@centrum.cz>
+    Copyright © 2018 Jonathan Hale <squareys@googlemail.com>
```
mosra: Weren't the shaders also part of the original repositories? If so, would be nice to mention the corresponding people here to avoid frowny faces ;)

src/Magnum/ImGuiIntegration/Integration.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+ImGuiShader::ImGuiShader() {
+    #ifdef MAGNUM_IMGUIINTEGRATION_BUILD_STATIC
+    /* Import resources on static build, if not already */
+    if(!Utility::Resource::hasGroup("MagnumImGuiIntegrationShaders"))
+        importShaderResources();
+    #endif
+    Utility::Resource rs("MagnumImGuiIntegrationShaders");
+
+#ifndef MAGNUM_TARGET_GLES
```
mosra: Indent the preprocessor as well, plz ;)

src/Magnum/ImGuiIntegration/Integration.cpp
```diff
@@ -0,0 +1,305 @@
...
+        frag.addSource({"#define NEW_GLSL\n"});
+    }
+
+    // Enable explicit attribute locations in shader if they are supported, otherwise bind locations directly
+    if(GL::Context::current().isExtensionSupported<GL::Extensions::ARB::explicit_attrib_location>(version)) {
+        vert.addSource({"#define EXPLICIT_ATTRIB_LOCATION\n"});
+    } else {
+        bindAttributeLocation(Position::Location, "Position");
+        bindAttributeLocation(TextureCoordinates::Location, "UV");
+        bindAttributeLocation(Color::Location, "Color");
```
mosra: At first I thought imgui requires the variables to be named like this, but apparently not. In that case, can these follow the naming scheme used in other shaders (in particular, lowercase first letter)? Thank you!

src/Magnum/ImGuiIntegration/Integration.h
Squareys: Oh, yeah, I think I fixed that _somewhere_ but aparently missed all other places :D

package/ci/appveyor.yml
Squareys: In this case it would almost make sense to just bundle it in external 🙈 (As submodule would be best, but that's non-trivial to use as non-git user... but if well documented...)

modules/FindMagnumIntegration.cmake
Squareys: Ah, right, now I remember why we had a discussion about this once :D

Squareys: [Documentation](https://cmake.org/cmake/help/v3.0/command/find_package.html) says "REQUIRED stops processing on error message", hence manually outputting an error message would make it stop if `find_package(MagnumIntegration REQUIRED ImGui)` whereas OPTIONAL would print the error message but continue processing...?

Squareys: OTOH, the error message will be output by `find_package(ImGui)` already... right? So it will stop processing there, I'd expect...

src/Magnum/ImGuiIntegration/Integration.cpp
Squareys: Yes, will do! Forgot about that 👍

modules/FindMagnumIntegration.cmake
mosra: Not sure, probably not. We need to revisit this later, maybe we need to add `REQUIRED` based on value of [MagnumIntegration_FIND_REQUIRED_ImGui](https://cmake.org/cmake/help/latest/command/find_package.html#id6) ... but that then needs to be done for *everything*.

package/ci/appveyor.yml
mosra: Well, I hope the changes are not so drastic (and I assume it will get more stable and better backward-compatible with upcoming versions). This is similar to OVR integration, no? There it was also recommending a version but it usually worked with newer ones as well.

Squareys:
@mosra Ready for review!

Open questions (/for follow up):
 - vcpkg requires support for linking imgui as lib rather than compiling in the code
 - Supported imgui version is mentioned nowhere
 - `REQUIRED` in FindMagnumIntegration

src/Magnum/ImGuiIntegration/Integration.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    CORRADE_INTERNAL_ASSERT_OUTPUT(GL::Shader::compile({vert, frag}));
+
+    attachShaders({vert, frag});
+
+    CORRADE_INTERNAL_ASSERT_OUTPUT(link());
+
+    _projMatrixUniform = uniformLocation("projectionMatrix");
+
+    setUniform(uniformLocation("colorTexture"), TextureLayer);
```
mosra: I have a feeling that the two lines above are also wrapped in some `if()` for the builtin shaders. I'll look into this.

src/Magnum/ImGuiIntegration/Test/IntegrationTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+    ImVec2 imVec2{1.1f, 1.2f};
+    Vector2 vec2(1.1f, 1.2f);
+
+    CORRADE_COMPARE(Vector2(imVec2), vec2);
+
+    ImVec2 c(vec2);
+    CORRADE_COMPARE(c.x, vec2.x());
+    CORRADE_COMPARE(c.y, vec2.y());
+}
+void IntegrationTest::vector4() {
```
mosra: Empty line before :trollface:

modules/FindImGui.cmake
```diff
@@ -0,0 +1,82 @@
...
+#   THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#   DEALINGS IN THE SOFTWARE.
+#
+
+list(FIND ImGui_FIND_COMPONENTS "Sources" _INDEX)
+if(_INDEX GREATER -1)
+    set(_FIND_SOURCES TRUE)
+endif()
```
mosra: Hmm. For a seamless vcpkg compatibility the presence of either the sources or a static lib should be abstracted away from the user. In case of vcpkg it's just `find_package(imgui CONFIG)`, right? Then I would do:

- the "canonical" way to do the above is a "loop switch", i.e. `foreach(_component ${ImGui_FIND_COMPONENTS})`. Then you set `ImGui_Sources_FOUND` accordingly and pass `HANDLE_COMPONENTS` to FPHSA to make it handle the components correctly. Also it's not a problem if a `ImGui::ImGui` or `ImGui::Sources` target is created before the FPHSA call.
- if `find_package(imgui CONFIG)` succeeds, then `ImGui::ImGui` would have just [imgui::imgui](https://github.com/Microsoft/vcpkg/blob/master/ports/imgui/CMakeLists.txt) in its `INTERFACE_LINK_IBRARIES`, plus, if requested, an (otherwise empty) `ImGui::Sources` target, which has `ImGui::ImGui` in its `INTERFACE_LINK_IBRARIES`
- if it fails, then it tries to find the include dir and adds that to `ImGui::ImGui` `INTERFACE_INCLUDE_DIRECTORIES`
- and for sources, if requested, it finds the three `*.cpp` files and creates a new `ImGui::ImGui` target with `INTERFACE_SOURCES` set to these and `ImGui::ImGui` in its `INTERFACE_LINK_LIBRARIES` (and *not* the other way around)

As far as I can see, the only user of `ImGui::Sources` would be `ImGuiIntegration`, which also sets up correct `IMGUI_API` depending on whether it is built dynamically or not. To its `INTERFACE` it would expose only `ImGui::ImGui`, not `ImGui::Sources`.

Hopefully I didn't miss anything blatantly obvious here ... uh, how should it behave regarding dllimport/dllexport if building the integration as dynamic and the vcpkg lib is static? Is that a valid scenario? Maybe in that case add the static lib also to `INTERFACE` link libraries?

Squareys: Yes, that's how I had it implemented until I realised:

What if the user builds with imgui sources and then uses vcpkg for his project? Then there would be duplicate symbols. Same for the other way around.

`IMGUI_API` is already being set up correcty, see [visibility.h](https://github.com/mosra/magnum-integration/pull/33/files#diff-77a8c1458c9c3ca37059a14f80fa00ac).

I guess, we'd put in a big warning of "make sure to link imgui consistently either with vcpkg or IMGUI_DIR (and sources)"

Which all made me decide to not support vcpkg for now, but maybe in a second step.

> the integration as dynamic and the vcpkg lib is static?

The vcpkg lib is dynamic if vcpkg is currently building integration with a dynamic triplet, if it's building static, imgui will be static aswell. Which, yes, means that the symbols will not be present and we would have to somehow merge imgui.lib into MagnumImGuiIntegration.lib using libtool or something. Or would I have the find module still link imgui.lib via `INTERFACE_LINK_LIBRARIES`?

I think assuming that if we found imgui through the package config, MagnumImGuiIntegration will have been build with the package config/precompiled version also should be a more or less safe bet. It's also the most user friendly...

mosra: > What if the user builds with imgui sources and then uses vcpkg for his project? [...] Same for the other way around.

How common is that? I don't think people are doing this. The other way around I don't see a problem either: it'll find the headers and link to just that, no duplicate symbols.

> we would have to somehow merge imgui.lib into MagnumImGuiIntegration.lib 

No, just have it in `INTERFACE_LINK_LIBRARIES` like I said above.

> I think assuming that if we found imgui through the package config, MagnumImGuiIntegration will have been build with the package config/precompiled version 

Yes. I don't see a value in supporting the very theoretical case of using magnum from outside of vcpkg together with vcpkg pacakges. You'd run into the same problems when combining e.g. PngImporter built outside of vcpkg against some external libPNG and then using that with libPNG from vcpkg.

src/Magnum/ImGuiIntegration/Integration.cpp
Squareys: So after looking into this, what do you say? (I am not sure whether I understand what this is about, btw...)

modules/FindImGui.cmake
Squareys: True, alright, will apply changes as suggested.

Squareys: > if find_package(imgui CONFIG) succeeds, then ImGui::ImGui would have just imgui::imgui in its INTERFACE_LINK_IBRARIES, plus, if requested, an (otherwise empty) ImGui::Sources target, which has ImGui::ImGui in its INTERFACE_LINK_IBRARIES

Uhm. But the `ImGui::Sources` would _not_ be in the `INTERFACE_LINK_LIBRARIES`, right? That wouldn't make sense as the symbols would exist twice then, once from the sources and once from the libs in the `imgui::imgui` target.

Squareys: Ah, I see now. The ImGui::Sources is supposed to be empty :)

modules/FindImGui.cmake
```diff
@@ -0,0 +1,101 @@
...
+                list(APPEND ImGui_SOURCES ${ImGui_${_file}_SOURCE})
+            endforeach()
+
+            add_library(ImGui::Sources INTERFACE IMPORTED)
+            set_property(TARGET ImGui::Sources APPEND PROPERTY
+                INTERFACE_SOURCES "${ImGui_SOURCES}")
+            set_property(TARGET ImGui::Sources APPEND PROPERTY
+                INTERFACE_LINK_LIBRARIES ImGui::ImGui)
+
+            set(ImGui_Sources_FOUND ${ImGui_SOURCES})
```
Squareys: @mosra How would I go about this? I though about a couple of if/else to set to TRUE or FALSE, but that sounds cumbersome. Maybe check the list length or something?

src/Magnum/ImGuiIntegration/Integration.cpp
mosra: Oh! I'll look into this during the merge, I meant.

modules/FindImGui.cmake
mosra: Set it to `ON` by default, then in the foreach above set it to `OFF` if `NOT ImGui_${_file}_SOURCE`.

Besides that, `if(NOT TARGET)` should wrap only the target addition, not the whole thing.

modules/FindImGui.cmake
```diff
@@ -0,0 +1,101 @@
...
+#   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+#   THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#   DEALINGS IN THE SOFTWARE.
+#
+
+list(FIND ImGui_FIND_COMPONENTS "Sources" _INDEX)
+if(_INDEX GREATER -1)
+    set(_FIND_SOURCES TRUE)
```
mosra: This is a complete leftover now, right?

src/Magnum/ImGuiIntegration/Integration.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    drawData->ScaleClipRects(io.DisplayFramebufferScale);
+
+    const Matrix4 projection{
+        {2.0f/io.DisplaySize.x, 0.0f, 0.0f, 0.0f},
+        {0.0f, -2.0f/io.DisplaySize.y, 0.0f, 0.0f},
+        {0.0f, 0.0f, -1.0f, 0.0f},
+        {-1.0f, 1.0f, 0.0f, 1.0f},
+    };
+    _shader.setProjectionMatrix(projection);
```
mosra: Not really sure, but maybe composing the matrix from primitive operations ("move origin to top left", "switch to Y down", "projection the size of the display") would make it appear a bit more self-documenting?

modules/FindImGui.cmake
mosra: Yes, empty, except for having ImGui::ImGui in its `INTERFACE_LINK_LIBRARIES`.

src/Magnum/ImGuiIntegration/Widgets.h
```diff
@@ -0,0 +1,66 @@
...
+#include <Magnum/Math/Vector4.h>
+#include <Magnum/Math/Color.h>
+#include <Magnum/GL/Texture.h>
+
+#include "Magnum/ImGuiIntegration/visibility.h"
+#include "Magnum/ImGuiIntegration/Conversion.h"
+
+#include <imgui.h>
+
+namespace Magnum { namespace ImGuiIntegration { namespace MgImGui {
```
mosra: Oh, haha, somehow forgot to comment on this namespace (wasn't there a discussion about the name before?). 

What about having `Image()` directly in `ImGuiIntegration`? (also why not `image()`, lowercase?). Another idea is have the namespace named `Widgets`, but since here is just a single function right now, I would place it directly inside the parent namespace.

src/Magnum/ImGuiIntegration/Widgets.h
```diff
@@ -0,0 +1,66 @@
...
+@param size Widget size
+@param uv0 Min of source range on the texture
+@param uv1 Max of source range on the texture
+@param tintColor Tint color, default @cpp 0xffffffff_rgbaf @ce
+@param borderColor Border color, default @cpp 0x00000000_rgbaf @ce
+*/
+void Image(GL::Texture2D& texture, const Vector2& size,
+    const Vector2& uv0=Vector2{}, const Vector2& uv1=Vector2{1.0f},
+    const Color4& tintColor=Color4{1.0f},
+    const Color4& borderColor=Color4{0.0f})
```
mosra: Spaces around `=` (we're not in Python here ;))

src/Magnum/ImGuiIntegration/Integration.h
```diff
@@ -0,0 +1,211 @@
...
+         * @brief Global context instance
+         *
+         * Expects that an instance has been created.
+         */
+        static Context& get();
+
+        /**
+         * @brief Start a new frame
+         * @param winSize Window size
+         * @param viewportSize Viewport size
```
mosra: Can youvertically align the parameter descriptions like it's done in Magnum itself? Preferrably to some tab-stop. Similarly elsewhere and in `Widgets.h`. Thanks!

mosra:
For the CI failures, it looks like you forgot to add the file in `doc/snippets` ;)

modules/FindImGui.cmake
Squareys: Yup 👍

src/Magnum/ImGuiIntegration/Widgets.h
Squareys: I commented on this only via gitter. My arguments were:

 - `ImGui::Image` would be best, as it would match `ImGui::Image`, but just overload the texture parameter
 - That doesn't work, though, as otherwise I would have to `::` prefix all ImGui calls, which looks pretty ugly
 - Hence I though to somewhat still match the style of ImGui calls and to make those easily distinguishable, `MgImGui::Image` would look nice and that same namespace could contain other future ImGui widgets

Squareys:
- [ ] there's some dllexport/dllimport issue on MinGW. Not sure why is that?
 - [ ] the Emscripten (at least WebGL 2) build should be working too (it was working in the original repos)
 - [ ] same for iOS ES3 I think
 - [ ] doc/namespaces.dox is missing the license info (i'm sure i mentioned this before somewhere, but can't find where)
 - [x] the comments below

src/Magnum/ImGuiIntegration/Widgets.h
mosra: > as otherwise I would have to :: prefix all ImGui calls

but only in the code of the integration itself and only for `Image` calls, no? That could be still acceptable I think. 

I would go with `ImGuiIntegration::image()`, since that's how we name functions here (similarly, we also don't use `snail_case` for APIs that extend / emulate STL, so why emulate ImGui naming). Or wrap it in `ImGuiIntegration::Widgets::image()`, but that's too much of a colon cancer for my taste, especially considering there's just a single function right now. `Magnum::Ui` also doesn't have any `Widgets` subnamespace either.

modules/FindImGui.cmake
```diff
@@ -0,0 +1,107 @@
...
+        endif()
+
+        foreach(_file widgets draw demo)
+            find_file(ImGui_${_file}_SOURCE NAMES imgui_${_file}.cpp
+                HINTS ${IMGUI_DIR})
+            list(APPEND ImGui_SOURCES ${ImGui_${_file}_SOURCE})
+            if(NOT ImGui_${_file}_SOURCE)
+                set(ImGui_Sources_FOUND FALSE)
+            endif()
+        endforeach()
```
mosra: Compiling ImGui sources as part of the client app will cause lots of warnings emitted on GCC, this could be (I hope) a solution:

```cmake
# Clang-specific flags
if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR (CMAKE_CXX_COMPILER_ID MATCHES "(Apple)?Clang" AND NOT CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC") OR CORRADE_TARGET_EMSCRIPTEN)
    set_property(FILE ${ImGui_${_file}_SOURCE} APPEND PROPERTY COMPILE_FLAGS 
        "-Wno-zero-as-nullptr-constant")
endif()

# GCC-specific flags
if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
    set_property(FILE ${ImGui_${_file}_SOURCE} APPEND PROPERTY COMPILE_FLAGS 
        "-Wno-double-promotion")
endif()
```

Squareys: It appears that `FILE` is not a valid scope for `set_property`. Full error:
```
CMake Error at modules/FindImGui.cmake:95 (set_property):
  set_property given invalid scope FILE.  Valid scopes are GLOBAL, DIRECTORY,
  TARGET, SOURCE, TEST, CACHE, INSTALL.
Call Stack (most recent call first):
  src/Magnum/ImGuiIntegration/CMakeLists.txt:30 (find_package)
```

Squareys: It was `set_property(SOURCE ... APPEND_STRING ...`, works, but still spews out a lot of warnings...

Squareys: Ah, wait, my bad

Squareys: I had to use spaces in front of the flag now with `APPEND_STRING`, obviously. The nullptr flag is spelled `null-pointer` and apart from that I also added `-Wno-old-style-cast` on GNU and clang and now it's silent! 👏 👏

modules/FindImGui.cmake
```diff
@@ -0,0 +1,107 @@
...
+#   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+#   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+#   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+#   THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#   DEALINGS IN THE SOFTWARE.
+#
+
+find_package(imgui CONFIG)
```
mosra: This should be `QUIET`, otherwise CMake will in most cases print this:

```
CMake Warning at modules/FindImGui.cmake:42 (find_package):
  Could not find a package configuration file provided by "imgui" with any of
  the following names:
    imguiConfig.cmake
    imgui-config.cmake
  Add the installation prefix of "imgui" to CMAKE_PREFIX_PATH or set
  "imgui_DIR" to a directory containing one of the above files.  If "imgui"
  provides a separate development package or SDK, be sure it has been
  installed.
```

modules/FindImGui.cmake
```diff
@@ -0,0 +1,125 @@
+#.rst:
+# Find ImGui
+# -------------
+#
+# Finds the ImGui library. This module defines:
+#
+#  ImGui_FOUND                - True if ImGui library is found
+#  ImGui::ImGui               - ImGui imported target (only on CMake >= 3.0,
+#   because interface targets are not supported on 2.8.12)
```
mosra: Comment not needed anymore :)

src/Magnum/ImGuiIntegration/Widgets.h
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+@brief Image widget displaying a @ref GL::Texture2D
+@param size         Widget size
+@param uv0          Min of source range on the texture
+@param uv1          Max of source range on the texture
+@param tintColor    Tint color, default @cpp 0xffffffff_rgbaf @ce
+@param borderColor  Border color, default @cpp 0x00000000_rgbaf @ce
+*/
+void Image(GL::Texture2D& texture, const Vector2& size,
```
mosra: Needs to be `inline`, otherwise you can get duplicate symbols when linking.

doc/snippets/ImGuiIntegration.cpp
```diff
@@ -0,0 +1,71 @@
...
+using namespace Magnum;
+
+int main() {
+#ifdef __clang__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wuninitialized"
+#elif defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
```
mosra: Not really needed here (I want to solve this in a more flexible way later, so the less of this cruft the better ;))

doc/snippets/ImGuiIntegration.cpp
```diff
@@ -0,0 +1,71 @@
...
+/* [Context-usage] */
+ImGuiIntegration::Context context;
+
+// ...
+
+ImGuiIntegration::Context::get().newFrame({640, 480}, {640, 480});
+/* [Context-usage] */
+
+/* [Context-usage-per-frame] */
+GL::Renderer::enable(GL::Renderer::Feature::Blending);
```
mosra: Maybe also enable blending only for the draw call?

modules/FindImGui.cmake
```diff
@@ -0,0 +1,130 @@
...
+                set(ImGui_Sources_FOUND FALSE)
+                continue()
+            endif()
+
+            # Hide warnings from imgui source files
+
+            set_property(SOURCE ${ImGui_${_file}_SOURCE} APPEND PROPERTY COMPILE_DEFINITIONS
+                "IMGUI_USER_CONFIG=\"Magnum/ImGuiIntegration/visibility.h\"")
+
+            # Clang-specific flags
```
mosra: `GCC- and Clang-specific` (sorry, my snippet was wrong). Also, maybe above a comment on what the IMGUI_USER_CONFIG does?

modules/FindMagnumIntegration.cmake
mosra: Can you remove the `REQUIRED`, then?

package/ci/appveyor.yml
```diff
@@ -60,11 +60,17 @@ install:
...
 - IF "%TARGET%" == "desktop" IF "%COMPILER%" == "mingw" pip install codecov
+- IF "%TARGET_GLES2%" == "ON" set TARGET_GLES3=OFF
+- IF "%TARGET_GLES2%" == "OFF" set TARGET_GLES3=ON
 
 # GLM
 - IF NOT EXIST %APPVEYOR_BUILD_FOLDER%\0.9.9.0.zip appveyor DownloadFile https://github.com/g-truc/glm/archive/0.9.9.0.zip
 - 7z x 0.9.9.0.zip -o%APPVEYOR_BUILD_FOLDER%\deps && ren %APPVEYOR_BUILD_FOLDER%\deps\glm-0.9.9.0 glm
 
+# ImGui
+- IF NOT EXIST %APPVEYOR_BUILD_FOLDER%\imgui-1.65.zip appveyor DownloadFile https://github.com/ocornut/imgui/archive/65dac021712f99ba954392734ceafe43392efe59.zip -FileName imgui.zip
```
mosra: should say `IF NOT EXIST %APPVEYOR_BUILD_FOLDER%\imgui.zip`, I think?

package/ci/travis.yml
```diff
@@ -207,6 +209,9 @@ install:
 - if [ "$TRAVIS_OS_NAME" == "linux" ] && [ ! "$TARGET" == "android" ]; then ./package/ci/travis-dependencies.sh; fi
 # TODO: enable on macOS once HOMEBREW_NO_AUTO_UPDATE=1 brew install dartsim is fixed
 
+# ImGui
+- cd $HOME ; wget -nc --no-check-certificate -O imgui.tar.gz https://github.com/ocornut/imgui/archive/65dac021712f99ba954392734ceafe43392efe59.tar.gz && mkdir -p imgui && cd imgui && tar --strip-components=1 -xzf ../imgui.tar.gz && cd $TRAVIS_BUILD_DIR
```
mosra: Add that to `cache:` above? :)

src/Magnum/ImGuiIntegration/Integration.cpp
```diff
@@ -0,0 +1,259 @@
...
+    _instance = this;
+}
+
+Context::~Context() {
+    /* Not initialized/initialization failed */
+    if(!_instance) return;
+
+    ImGui::DestroyContext();
+
+    _instance = nullptr;
```
mosra: `CORRADE_INTERNAL_ASSERT(_instance == this);` before to prevent accidental bugs later

src/Magnum/ImGuiIntegration/Integration.hpp
mosra: Still not fixed ;)

src/Magnum/ImGuiIntegration/Test/IntegrationGLTest.cpp
```diff
@@ -0,0 +1,288 @@
...
+        Context c;
+        MAGNUM_VERIFY_NO_GL_ERROR();
+
+        std::ostringstream out;
+        Error redirectError{&out};
+        Context c2;
+
+        CORRADE_COMPARE(out.str(),
+                "ImGuiIntegration::Context::Context(): Context already created.\n");
+    }
```
mosra: Is the scope needed, since there's no check after?

src/Magnum/ImGuiIntegration/Test/IntegrationGLTest.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+void IntegrationGLTest::textInput() {
+    Context c;
+
+    TextInputEvent textEvent{{"abc"}};
+    c.handleTextInputEvent(textEvent);
+
+    ImWchar expected[4]{'a', 'b', 'c', '\0'};
+    CORRADE_COMPARE_AS(Containers::arrayView(ImGui::GetIO().InputCharacters, 4),
+            Containers::arrayView(expected, 4), TestSuite::Compare::Container);
```
mosra: should work even without the `, 4` at the end

doc/snippets/ImGuiIntegration.cpp
Squareys: I assume you mean disabling it again after

williamjcm:
Would it be possible for the official integration to have a way to load custom fonts ?

I currently have a customised version of nnarain's integration, and I had to edit it to add code to load my own fonts.

I went to [this line](https://github.com/nnarain/magnum-imgui/blob/256968783121306c65ddfb5c28906140f2ccf947/src/MagnumImGui.cpp#L74), and added the following before it (I load the same font multiple times, just at different sizes to fit the users' eventual needs):

```cpp
io.Fonts->AddFontFromFileTTF("aller.ttf",
                             16.0f,
                             nullptr,
                             ranges.Data);
io.Fonts->AddFontFromFileTTF("aller.ttf",
                             18.0f,
                             nullptr,
                             ranges.Data);
io.Fonts->AddFontFromFileTTF("aller.ttf",
                             20.0f,
                             nullptr,
                             ranges.Data);
io.Fonts->AddFontFromFileTTF("aller.ttf",
                             22.0f,
                             nullptr,
                             ranges.Data);
```

As you can see, I load the same font multiple times at different sizes (to fit users' needs).

I don't specify a custom font config, but I use a custom character range (the `ranges` variable) to add support for all the glyphs in the font. Therefore, having both as optional parameters in the integration could be a possibility.

Here's the ImGui documentation about fonts: https://github.com/ocornut/imgui/blob/67775448556d8dba304657719829f1ab768b3c97/imgui.cpp#L764-L838 and https://github.com/ocornut/imgui/blob/master/misc/fonts/README.txt

Squareys:
@williamjcm You can still add those lines after calling the `Magnum::ImGuiIntegration::Context` constructor in your user code. If you use `ImGui::GetIO()`, you will retrieve the same IO struct that is being used by Magnum.

Otherwise a more magnum-like solution (? @mosra ) would probably be to allow ImGui to use `Magnum::Text::AbstractFont` somehow, but since this is likely quite a bit of work, it will definitely not be part of this PR.

williamjcm:
The code that generates the font texture data is in the integration, however.

Squareys:
Oh! True, sorry, thanks for the hint!

@mosra Maybe it would be best to add a dependency on AbstractFont here then and let that generate the glyph texture instead? I wouldn't expect the textures to be compatible, though, so maybe a "generateFonts" or something as a second step after Context initialization. I wonder how we could make that "unforgetible", i.e. an optional second step if no custom font needs to be loaded 🤔

williamjcm:
I like that idea.

It would give more control to the integration's users who want it, while still giving a "I don't care" option to those who just want to get ImGui up and running.

mosra:
I merged the PR with the following changes:

- The headers are renamed to `Context.h` for the `Context` class and `Integration.h` for the math integration, to be more in-line with the naming elsewhere
- The above proposal for custom fonts is integrated
- The API had to be changed for DPI awareness, so instead of `nextFrame()` taking window / framebuffer size, a tuple of UI size, window size and framebuffer size is passed to the constructor and there's a new `relayout()` function to handle window size or pixel density changes. This is basically equivalent to how it's done in the `Ui` library, so you can just reuse the logic from there. The docs (which I'll be uploading later) explain the interactions of DPI awareness with custom fonts.
- The singleton and `get()` is no more, instead you are expected to handle a singleton yourself, *if needed* -- when working on the example I realized that the singleton was not needed at all, since the only place where the `nextFrame()` / `drawFrame()` and event functions were called was in the main app code, which also is the owner of the `Context` instance. This allowed me to make it move-aware, have a `NoCreate` constructor and support multiple contexts, and also removing all restrictive singleton-related assertions from the internals.
- The `image()` widget now has the UV range represented as `Range2D` instead of a pair of vectors.
- There's also conversion between `Color3` and `ImColor`, to allow things like `ImColor(0xff3366_rgbf)` (and avoid crazy implicit-conversion-caused bugs, more in the docs)
- Mouse position is registered also for mouse move / scroll events, for some reason it was done only for mouse move event, which would cause issues on touchscreens (which have no move events reported)
- The Context class is now marked as experimental, since I'm not really sure how to implement a Vulkan backend yet and it might involve some renaming, so better safe than sorry ;)

I did not bother with port to ES2 / WebGL 1 yet since that would delay the merge *by another year*, will do that later. I'm now working on the example, will upload the docs once I have the example updated and merged as well.

Feedback and bugreports very appreciated :)

PR #34, opened 2019-01-08
BulletIntegration: fix Matrix3x3 conversion along with unit tests

xqms:
Bullet matrices are row-major, Magnum matrices are column-major.
The `Matrix3x3` unit test did not account for this and tested the wrong behavior.
The test for `Matrix4x4` seems fine, haven't looked in detail, though.

This probably hasn't been noticed before because `BulletIntegration::MotionState` uses the `Matrix3x3` conversion only in the Magnum -> Bullet direction, and this is typically only done once during `btRigidBody` initialization. The bullet example in magnum-examples uses identity rotation at initialization...

### DIFF
--- a/src/Magnum/BulletIntegration/Integration.h
+++ b/src/Magnum/BulletIntegration/Integration.h
@@ -52,15 +52,15 @@ template<> struct VectorConverter<3, btScalar, btVector3> {
 
 template<> struct RectangularMatrixConverter<3, 3, btScalar, btMatrix3x3> {
     static RectangularMatrix<3, 3, Float> from(const btMatrix3x3& other) {
-        return {Vector<3, Float>(other[0]),
-                Vector<3, Float>(other[1]),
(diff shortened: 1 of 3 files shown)


mosra:
Hi, thanks a lot for reporting and fixing this!

Interesting (and frightening) that this hasn't been noticed yet, huh. Looking into Bullet sources at how the Matrix4 conversion is done, yeah this was definitely wrong. @Squareys were you aware of this?

Merged your commit in 67059e523b1f61ce0ce6f117380736b64bde7de8 (without the debug prints) and then, in 13356bf68e78ebef2e4b3ebfa375d10c641ce9ee, to be really sure, updated the 3x3 test to now also convert the matrix to a quaternion and check that it indeed produces a correct value.

xqms:
Ah, sorry, the debug stuff should not have been in there, that slipped through. Thanks for checking and merging :-)

PR #38, opened 2019-02-17
ImGuiIntegration: add imageButton()

williamjcm:
Also adds the corresponding test in ImGuiWidgetsGLTest.

Since I tested with ImGui 1.68 WIP, which [apparently needs some time to display stuff](https://github.com/ocornut/imgui/commit/3c07ec6a6126fb6b98523a9685d1f0f78ca3c40c), I had to edit both ImGuiWidgetsGLTest and ImGuiContextGLTest to add some delay between the dummy first frame and the actual test frames.

Review and feedback are welcome!

### DIFF
--- a/src/Magnum/ImGuiIntegration/Widgets.h
+++ b/src/Magnum/ImGuiIntegration/Widgets.h
@@ -29,7 +29,7 @@
 */
 
 /** @file
- * @brief Functions @ref Magnum::ImGuiIntegration::image(), @ref Magnum::ImGuiIntegration::imageButton()
+ * @brief Function @ref Magnum::ImGuiIntegration::image(), @ref Magnum::ImGuiIntegration::imageButton()
  */
 
 #include <imgui.h>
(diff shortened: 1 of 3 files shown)


src/Magnum/ImGuiIntegration/Widgets.h
```diff
@@ -29,7 +29,7 @@
 */
 
 /** @file
- * @brief Function @ref Magnum::ImGuiIntegration::image()
+ * @brief Functions @ref Magnum::ImGuiIntegration::image(), @ref Magnum::ImGuiIntegration::imageButton()
```
mosra: Keep the `Function`, since I already have that everywhere else (consistency).

src/Magnum/ImGuiIntegration/Widgets.h
```diff
@@ -58,6 +58,24 @@ inline void image(GL::Texture2D& texture, const Vector2& size,
...
+@param uvRange          UV range on the texture (covers the whole texture by default)
+@param framePadding     Frame padding, negative values use the default frame padding.
+@param backgroundColor  Background color, default @cpp 0x00000000_rgbaf @ce
+@param tintColor        Tint color, default @cpp 0xffffffff_rgbaf @ce
+*/
+inline bool imageButton(GL::Texture2D& texture, const Vector2& size,
+                        const Range2D& uvRange = {{}, Vector2{1.0f}},
+                        const int& framePadding = -1,
+                        const Color4& backgroundColor = Color4{0.0f},
+                        const Color4& tintColor = Color4{1.0f})
```
mosra: Indent just by four spaces, like above, makes it possible to fit more things on the screen. Also:

- can be just `backgroundColor = {}`.
- use just `int` instead of `const int&`, the reference is actually bigger than the int on 64bit platforms so it doesn't save anything

src/Magnum/ImGuiIntegration/Widgets.h
williamjcm: Got it.

src/Magnum/ImGuiIntegration/Widgets.h
williamjcm: Sorry about that. QtCreator forced its indentation settings, and I forgot to re-apply the indentation used by image().

mosra:
Merged as db791bed39d20caf301fa44e55bd000b0ec11ebf, c715e73bd81b11e00e95ec908c4fa1d791b6b0d8 and 62918a267a84d33e4ab7ed7b79fc5168a3a4921e. Thanks a lot! :+1:

PR #53, opened 2019-09-12
Allows building using Emscripten and 'WITH_BULLET' set

bowling-allie:
if 'EMSCRIPTEN' is on, 'find_package(Bullet)' and others are replaced
with 'target_compile_options(... -s USE_BULLET=1)' effectifely using
the emscripten-ports/Bullet library


bowling-allie:
the build might have failed because there were 4 jobs running?
one job downloaded the port(bullet), another other one didnt and errored when it saw an include
kinda doesnt make sense tho
builds on my machine using wsl Ubuntu 18.04.3 LTS

bowling-allie:
Ya, the emscripten-ports/bullet lib is from 2013. It's just too dang old for the tests

mosra:
Hi, sorry for the delays on my end -- 

just to give you an update: I'm in process of merging this, just wanted to expose the feature to `FindMagnumIntegration.cmake` as well so the `USE_BULLET` flag gets propagated also with separate installations and not only CMake subprojects. I have to run now but will have that done by tomorrow. And I'll have a look at mosra/toolchains#8 as well :)

bowling-allie:
Thank you!

mosra:
Sorry for the constant delays :) ... extended version merged as 4328706bd5a3de7d2242689227fdcf29fc80c076. Turns out the missing header in tests was present in two places and the other place is available consistently everywhere. Not sure why is it duplicated like that, but most importantly the tests pass also with this ages old version. (But I didn't enable it on the CI, I still prefer the new version more.)

The `-s USE_BULLET=1` flag unfortunately needs two features present in CMake 3.12 and 3.13 (e.g., when combined with `-s USE_WEBGL=2`, CMake would happily turn it into `-s USE_BULLET=1 USE_WEBGL=2` and, ... well. This is one point of totally unnecessary extreme suffering I hate about Emscripten (it could accept `-sUSE_BULLET=1` and everything would be fine again but it doesn't), hopefully this increased version requirement for this particular feature is not a problem for you.

Also added this to the docs (and you to credits). Thanks for the contribution! :)

PR #56, opened 2019-10-19
ImGuiIntegration: implement updating application cursor

mosra:
Depends on mosra/magnum#383. Putting into a PR because I'm not sure if I can wrap up the other one in time for a release.

Cc: @Melix19

### DIFF
--- a/src/Magnum/ImGuiIntegration/Context.hpp
+++ b/src/Magnum/ImGuiIntegration/Context.hpp
@@ -213,6 +213,63 @@ template<class TextInputEvent> bool Context::handleTextInputEvent(TextInputEvent
     return false;
 }
 
+/* GLFW doesn't have all cursor types, so employing a dirty SFINAE trick that
+   falls back to Arrow if given enum value is not available */
+namespace Implementation {
+
+#ifndef DOXYGEN_GENERATING_OUTPUT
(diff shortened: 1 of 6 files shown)


src/Magnum/ImGuiIntegration/Context.hpp
```diff
@@ -202,6 +202,63 @@ template<class TextInputEvent> bool Context::handleTextInputEvent(TextInputEvent
...
+
+}
+
+template<class Application> void Context::updateApplicationCursor(Application& application) {
+    /* Ensure we use the context we're linked to */
+    ImGui::SetCurrentContext(_context);
+
+    switch(ImGui::GetMouseCursor()) {
+        case ImGuiMouseCursor_TextInput:
+            application.setCursor(Application::Cursor::TextInput);
```
Melix19: *Application::Cursor::Text

Or maybe you prefer TextInput as name and this is a way to telling me? 👀

mosra: Argh, yes, when seeing that naming in ImGui it immediately looked better than CSS's `text` and SDL/GLFW's `IBeam`. But of course I forgot to mention that when doing the final review on your PR :see_no_evil:

PR #58, opened 2019-11-06
DartIntegration: use EigenIntegration and small check for NaN values

costashatz:
This PR adds the following to the DartIntegration:

- uses EigenIntegration when needed
- adds a small check for NaN values (so that Magnum asserts do not break the flow of the program)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,12 +35,10 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/modules/" ${CMAKE_MODULE_PATH})
 find_package(Magnum REQUIRED)
 
-include(CMakeDependentOption)
-
 # Parts of the library
 option(WITH_BULLET "Build BulletIntegration library" OFF)
 option(WITH_DART "Build DartIntegration library" OFF)
-cmake_dependent_option(WITH_EIGEN "Build EigenIntegration library" ON "WITH_DART" OFF)
+option(WITH_EIGEN "Build EigenIntegration library" OFF)
 option(WITH_GLM "Build GlmIntegration library" OFF)
 option(WITH_IMGUI "Build ImGuiIntegration library" OFF)
 option(WITH_OVR "Build OvrIntegration library" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,10 +35,12 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/modules/" ${CMAKE_MODULE_PATH})
 find_package(Magnum REQUIRED)
 
+include(CMakeDependentOption)
+
 # Parts of the library
 option(WITH_BULLET "Build BulletIntegration library" OFF)
 option(WITH_DART "Build DartIntegration library" OFF)
-option(WITH_EIGEN "Build EigenIntegration library" OFF)
+cmake_dependent_option(WITH_EIGEN "Build EigenIntegration library" ON "WITH_DART" OFF)
 option(WITH_GLM "Build GlmIntegration library" OFF)
 option(WITH_IMGUI "Build ImGuiIntegration library" OFF)
 option(WITH_OVR "Build OvrIntegration library" OFF)
--- a/src/Magnum/DartIntegration/CMakeLists.txt
+++ b/src/Magnum/DartIntegration/CMakeLists.txt
@@ -25,6 +25,7 @@
 #
 
 find_package(Magnum REQUIRED GL SceneGraph MeshTools Primitives)
+find_package(MagnumIntegration REQUIRED Eigen)
 find_package(DART 6.0.0 CONFIG REQUIRED)
 
 if(BUILD_STATIC)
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -35,10 +35,12 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/modules/" ${CMAKE_MODULE_PATH})
 find_package(Magnum REQUIRED)
 
+include(CMakeDependentOption)
+
 # Parts of the library
 option(WITH_BULLET "Build BulletIntegration library" OFF)
 option(WITH_DART "Build DartIntegration library" OFF)
-option(WITH_EIGEN "Build EigenIntegration library" OFF)
+cmake_dependent_option(WITH_EIGEN "Build EigenIntegration library" ON "WITH_DART" OFF)
```
mosra: Not needed, like I said there's just the header dependency ;)

src/Magnum/DartIntegration/CMakeLists.txt
```diff
@@ -68,6 +69,7 @@ target_link_libraries(MagnumDartIntegration
     Magnum::SceneGraph
     Magnum::Primitives
     Magnum::MeshTools
+    MagnumIntegration::Eigen
```
mosra: Likewise, those two are not needed.

src/Magnum/DartIntegration/Object.cpp
```diff
@@ -62,32 +64,30 @@ Object& Object::update(Trade::AbstractImporter* importer) {
...
-        axis = axis.array() / std::sqrt(1-quat.w()*quat.w());
-        axis.normalize();
-    }
-    else
-        axis(0) = 1.;
+        trans = Matrix4(Matrix4d(_node->getRelativeTransform()));
 
-    Eigen::Vector3d T = trans->translation();
+    /* Check if any value is NaN */
+    if(Math::isNan(trans[0]).any() || Math::isNan(trans[1]).any() || Math::isNan(trans[2]).any() || Math::isNan(trans[3]).any()) {
```
mosra:
```suggestion
    if(Math::isNan(trans.toVector()).any()) {
```

src/Magnum/DartIntegration/Object.cpp
```diff
@@ -62,32 +64,30 @@ Object& Object::update(Trade::AbstractImporter* importer) {
...
+    if(Math::abs(Float(angle))<=1e-5f) {
+        axis = {1.f, 0.f, 0.f};
+    }
+    axis = axis.normalized();
+    Vector3 t = trans.translation();
 
     /* Pass it to Magnum */
     _transformation.resetTransformation()
-        .rotate(theta, u)
+        .rotate(angle, axis)
```
mosra: Why go through the pain of creating a quaternion and then again converting that to a transformation matrix inside `rotate()`? Wouldn't just the following work as well?

```cpp
_transformation.setTransformation(trans);
```

costashatz: `error: ‘class Magnum::SceneGraph::AbstractBasicTranslationRotation3D<float>’ has no member named ‘setTransformation’; did you mean ‘resetTransformation’?`

CMakeLists.txt
costashatz: But if we haven't installed EigenIntegration, how are we going to find the header?

src/Magnum/DartIntegration/Object.cpp
mosra: blargh, forgot about this :sweat_smile: ... okay then, keep it as-is

CMakeLists.txt
mosra: The header is only needed when compiling the DartIntegration, and it's right in the directory next to DartIntegration :)

src/Magnum/DartIntegration/Object.cpp
costashatz: I guess you should provide this function? What is the rationale of not providing it for AbstractFeature?

CMakeLists.txt
costashatz: Ahhh ok.. That's what you meant! OK. Doing this..

src/Magnum/DartIntegration/Object.cpp
mosra: Because some transformations support only translation, or only translation/rotation, and the full transformation matrix could be too general for these.

... but I should provide `rotate()` taking a quat, definitely, adding that to my TODOs.

src/Magnum/DartIntegration/Object.cpp
```diff
@@ -62,32 +64,30 @@ Object& Object::update(Trade::AbstractImporter* importer) {
...
+    }
 
-    /* Convert it to axis-angle representation */
-    Math::Vector3<Float> t(T[0], T[1], T[2]);
-    Math::Vector3<Float> u(axis(0), axis(1), axis(2));
-    Rad theta(angle);
+    Quaternion quat = Quaternion::fromMatrix(trans.rotationScaling());
+    Vector3 axis = quat.axis();
+    Rad angle = quat.angle();
+    if(Math::abs(Float(angle))<=1e-5f) {
```
mosra:
```suggestion
    if(Math::abs(angle) <= 1e-5_radf) {
```

does this work? :)

costashatz: It is!

mosra:
Merged as f55a61664532999bf91904b75d312252413af949 and 95d46a3c234423a70f482335db709e5b045fa857, thank you!

PR #62, opened 2020-01-05
ImGuiIntegration: Update reference to Context::_texture after moving Context

Squareys:
Hi @mosra !

As mentioned on #61, the move constructor needs to update the pointer to `_texture`, otherwise rendering breaks because it uses the then deinitialized texture memory.

Best,
Jonathan

### DIFF
--- a/src/Magnum/ImGuiIntegration/Context.cpp
+++ b/src/Magnum/ImGuiIntegration/Context.cpp
@@ -130,6 +130,11 @@ Context::Context(NoCreateT) noexcept: _context{nullptr}, _shader{NoCreate}, _tex
 
 Context::Context(Context&& other) noexcept: _context{other._context}, _shader{std::move(other._shader)}, _texture{std::move(other._texture)}, _vertexBuffer{std::move(other._vertexBuffer)}, _indexBuffer{std::move(other._indexBuffer)}, _timeline{std::move(other._timeline)}, _mesh{std::move(other._mesh)}, _supersamplingRatio{other._supersamplingRatio}, _eventScaling{other._eventScaling} {
     other._context = nullptr;
+    /* Update the pointer to _texture */
+    ImGuiContext* current = ImGui::GetCurrentContext();
+    ImGui::SetCurrentContext(_context);
+    ImGui::GetIO().Fonts->SetTexID(reinterpret_cast<ImTextureID>(&_texture));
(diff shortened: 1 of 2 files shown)


src/Magnum/ImGuiIntegration/Context.cpp
```diff
@@ -130,6 +130,8 @@ Context::Context(NoCreateT) noexcept: _context{nullptr}, _shader{NoCreate}, _tex
 
 Context::Context(Context&& other) noexcept: _context{other._context}, _shader{std::move(other._shader)}, _texture{std::move(other._texture)}, _vertexBuffer{std::move(other._vertexBuffer)}, _indexBuffer{std::move(other._indexBuffer)}, _timeline{std::move(other._timeline)}, _mesh{std::move(other._mesh)}, _supersamplingRatio{other._supersamplingRatio}, _eventScaling{other._eventScaling} {
     other._context = nullptr;
+    /* Update the pointer to _texture */
+    ImGui::GetIO().Fonts->SetTexID(reinterpret_cast<ImTextureID>(&_texture));
```
mosra: Maybe this should do `ImGui::SetCurrentContext(_context);` first, to ensure we don't update some other context.

mosra:
Oh and a test for this would be vital -- here https://github.com/mosra/magnum-integration/blob/d4f60dfc5c0aa257f50e960d2904495fb8ae6258/src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp#L262 it needs to check that `&b.atlasTexture() == io.Fonts.TexID`. And also `&c.atlasTexture() == io.Fonts.TexID`, which will notify you that you need to do a similar change in the move assignment as well.

src/Magnum/ImGuiIntegration/Context.cpp
Squareys: And then set it back later?

Squareys:
@mosra I updated the PR to reflect suggested changes. Note that I am restoring the original imgui context, which is backwards compatible and also tested by the tests to be the case.

Squareys:
Awesome, thanks! 🙏

bowling-allie:
I'm so embarrassed, i can't believe i missed this!

mosra:
No problem, @Squareys missed a part of this also ... and I should be blamed for not noticing the issues in both PRs ;)

PR #63, opened 2020-01-05
ImGuiIntegration: Fix assertion when move assigning empty context

Squareys:
@mosra As discussed via gitter, here's the follow-up fix for #62 .

Best,
Jonathan.

### DIFF
--- a/src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp
+++ b/src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp
@@ -127,6 +127,7 @@ struct ContextGLTest: GL::OpenGLTester {
     void constructExistingContext();
     void constructExistingContextAddFont();
     void constructMove();
+    void moveAssignEmpty();
 
     void release();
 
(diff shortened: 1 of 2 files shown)


src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp
```diff
@@ -291,6 +293,19 @@ void ContextGLTest::constructMove() {
...
+    Context a{NoCreate};
+    Context b{{}};
+
+    /* Move from empty */
+    b = std::move(a);
+    CORRADE_COMPARE(b.context(), nullptr);
+
+    /* Move into empty */
+    b = std::move(a);
+    CORRADE_COMPARE(a.context(), nullptr);
```
mosra: Maybe add a third variant, moving empty into empty? It won't trigger anything in the current code but better safe than sorry :D

Squareys: Sure

PR #65, opened 2020-01-11
ImGuiIntegrationTest: ShowTestWindow will be removed in v1.75

bowling-allie:
ShowTestWindow was deprecated in v1.53 and will be removed in v1.75


mosra:
Thank you! 👍 I also made the corresponding change in examples, mosra/magnum-examples@ec09358995f24e14137a364f30cec354eada401d.

PR #66, opened 2020-01-20
Add usage hints to vector-math integrations.

JacksonCampolattaro:
This PR request adds short explanations of the functionality included in GlmIntegration and EigenIntegration. This should make it more obvious that both allow for casting to and from Magnum's native vector types. It should also make it more clear which headers must be included to take advantage of this functionality.

Short examples of the syntax are adapted from the blog post that explained the integration features when they were first introduced.


mosra:
Thank you! :) Merged as a4e3263754b5b2eadb35834b529ceced3767abe7, I just converted the code snippets to be checked for compile errors (which uncovered an accidentally reversed order of arguments in `rotation()`, yay!).

PR #72, opened 2020-06-26
ImGuiIntegration: allow hiding the cursor

pezcode:
Small change that allows hiding the cursor with `ImGui::SetMouseCursor(ImGuiMouseCursor_None)`. Also updated the test to not assume that `ImGuiMouseCursor_None` is unknown.


pezcode:
Oh dear, this exploded. I only tested this with GLFW, let me actually *run* the tests offline.

mosra:
Thanks! :)

PR #74, opened 2020-07-28
Eigen <-> StridedArrayView integration

Janos95:
This patch provides an arrayCast method converting between Eigen expressions and Corrade's StridedArrayView class. Not the most beautiful code in the world but should mostly work ;).

### DIFF
--- a/doc/snippets/EigenIntegration.cpp
+++ b/doc/snippets/EigenIntegration.cpp
@@ -69,17 +69,22 @@ static_cast<void>(d);
 float data[15];
 Containers::StridedArrayView2D<float> view{data, {3,5}};
 
-/* To avoid a copy make sure to provide the correct type or use auto. */
+/**
+To avoid a copy make sure to either use auto, or specify the correct type.
+In this case the correct type would be
+Eigen::Map<Eigen::Matrix<float, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>, Eigen::Unaligned, Eigen::Stride<Eigen::Dynamic, Eigen::Dynamic>>.
+*/
 auto map = EigenIntegration::arrayCast(view);
 
-/* Copy everything into Eigen Matrix. */
-Eigen::MatrixXf m = map;
+/* To actually copy the data, specify an Eigen Matrix as the declaration type. */
+Eigen::MatrixXf m = EigenIntegration::arrayCast(view);
 
 /* Get back a StridedArrayView onto the second row. */
 Containers::StridedArrayView1D<float> rowView = EigenIntegration::arrayCast(m.row(2));
 
 /* [IntegrationStridedArrayView] */
 static_cast<void>(rowView);
+static_cast<void>(map);
 }
 
 {
--- a/doc/snippets/EigenIntegration.cpp
+++ b/doc/snippets/EigenIntegration.cpp
@@ -65,6 +65,24 @@ static_cast<void>(d);
 }
 
 {
+/* [IntegrationStridedArrayView] */
+float data[15];
+Containers::StridedArrayView2D<float> view{data, {3,5}};
+
+/* To avoid a copy make sure to provide the correct type or use auto. */
+auto map = EigenIntegration::arrayCast(view);
+
+/* Copy everything into Eigen Matrix. */
+Eigen::MatrixXf m = map;
+
+/* Get back a StridedArrayView onto the second row. */
+Containers::StridedArrayView1D<float> rowView = EigenIntegration::arrayCast(m.row(2));
+
+/* [IntegrationStridedArrayView] */
+static_cast<void>(rowView);
+}
+
+{
 /* [GeometryIntegration] */
 auto a = Matrix3::translation({-1.5f, 0.3f})*
          Matrix3::rotation(25.0_degf)*
--- a/src/Magnum/EigenIntegration/Integration.h
+++ b/src/Magnum/EigenIntegration/Integration.h
@@ -370,15 +370,14 @@ template<class To, std::size_t size, class T> inline To cast(const Math::Vector<
     return Math::Implementation::VectorConverter<size, T, To>::to(from);
 }
 
-
 /**
 @brief Convert Corrades StridedArrayView2D to Eigens Dynamic Matrix Type.
 */
 template<class T> inline Eigen::Map<Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>, Eigen::Unaligned, Eigen::Stride<Eigen::Dynamic, Eigen::Dynamic>> arrayCast(const Containers::StridedArrayView2D<T>& from) {
     using MatrixT = Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
     using StrideT = Eigen::Stride<Eigen::Dynamic, Eigen::Dynamic>;
-    const Containers::StridedDimensions2D<std::size_t> size = from.size();
-    const Containers::StridedDimensions2D<std::ptrdiff_t> stride = from.stride();
+    const Containers::StridedDimensions<2, std::size_t> size = from.size();
+    const Containers::StridedDimensions<2, std::ptrdiff_t> stride = from.stride();
     return Eigen::Map<MatrixT, Eigen::Unaligned, StrideT>(reinterpret_cast<T*>(from.data()), size[0], size[1], StrideT(stride[0]/sizeof(T), stride[1]/sizeof(T)));
 }
 
@@ -400,7 +399,7 @@ to a one dimensional StridedArrayView.
 */
 template<class Derived> inline typename std::enable_if<(Eigen::internal::traits<Derived>::ColsAtCompileTime == 1) || (Eigen::internal::traits<Derived>::RowsAtCompileTime == 1), Containers::StridedArrayView1D<typename Derived::Scalar>>::type arrayCast(const Eigen::DenseCoeffsBase<Derived, Eigen::DirectWriteAccessors>& from) {
     using Scalar = typename Derived::Scalar;
-    return Containers::StridedArrayView1D<Scalar>({const_cast<Scalar*>(&from(0,0)), std::size_t(from.size()*from.innerStride())}, std::size_t(from.size()), std::size_t(from.innerStride())*sizeof(Scalar));
+    return Containers::StridedArrayView1D<Scalar>({const_cast<Scalar*>(&from(0,0)), ~std::size_t{}}, std::size_t(from.size()), std::ptrdiff_t(from.innerStride()*sizeof(Scalar)));
 }
 
 /**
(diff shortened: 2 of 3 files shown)


src/Magnum/EigenIntegration/Integration.h
```diff
@@ -369,6 +370,51 @@ template<class To, std::size_t size, class T> inline To cast(const Math::Vector<
...
+}
+
+/**
+@brief Convert an Eigen Expression to Corrades StridedArrayView2D
+Overload that takes care of any Eigen Expression that was not handled by @ref Magnum::EigenIntegration::arrayCast().
+@snippet EigenIntegration.cpp IntegrationStridedArrayView
+*/
+template<class Derived> inline typename std::enable_if<(Eigen::internal::traits<Derived>::ColsAtCompileTime == Eigen::Dynamic) && (Eigen::internal::traits<Derived>::RowsAtCompileTime == Eigen::Dynamic), Containers::StridedArrayView2D<typename Derived::Scalar>>::type arrayCast(const Eigen::DenseCoeffsBase<Derived, Eigen::DirectWriteAccessors>& from) {
+    using Scalar = typename Derived::Scalar;
+    /* we assume that the Eigen expression is a slice of some continuous piece of memory */
```
mosra, line 413: Err... not happy about this comment :sweat_smile: 

So if I do a `image.pixels<Color4>().flipped<0>()` (i.e., image pixels with Y flipped), pass it to Eigen to do whatever magic something, and then convert back to a `StridedArrayView2D`, what do I get? This is one of the things that's best to either have implement fully (so there are no surprises when one gets slightly advanced) or not at all.

Looking further, the `size` is being used only in the first argument, where Corrade needs it only for an assert. I think we could safely assume the Eigen memory has correct bounds, and sidestep the assert by passing `~std::size_t{}` instead of the `size`.

src/Magnum/EigenIntegration/Integration.h
```diff
@@ -369,6 +370,51 @@ template<class To, std::size_t size, class T> inline To cast(const Math::Vector<
...
+    std::size_t size = from.rows()*from.colStride() + from.cols()*from.rowStride();
+    return Containers::StridedArrayView2D<Scalar>({const_cast<Scalar*>(&from(0,0)), std::size_t(size)}, {std::size_t(from.rows()), std::size_t(from.cols())}, {Long(from.rowStride())*Long(sizeof(Scalar)), Long(from.colStride())*Long(sizeof(Scalar))});
```
mosra, line 415: All `Long`s should be `std::ptrdiff_t` here, otherwise you'd get warnings/errors on 32bit.

src/Magnum/EigenIntegration/Integration.h
```diff
@@ -369,6 +370,51 @@ template<class To, std::size_t size, class T> inline To cast(const Math::Vector<
...
+    using Scalar = typename Derived::Scalar;
+    return Containers::StridedArrayView1D<Scalar>({const_cast<Scalar*>(&from(0,0)), std::size_t(from.size()*from.innerStride())}, std::size_t(from.size()), std::size_t(from.innerStride())*sizeof(Scalar));
+}
+
+/**
+@brief Convert an Eigen Expression to Corrades StridedArrayView2D
+Overload that takes care of any Eigen Expression that was not handled by @ref Magnum::EigenIntegration::arrayCast().
+@snippet EigenIntegration.cpp IntegrationStridedArrayView
+*/
+template<class Derived> inline typename std::enable_if<(Eigen::internal::traits<Derived>::ColsAtCompileTime == Eigen::Dynamic) && (Eigen::internal::traits<Derived>::RowsAtCompileTime == Eigen::Dynamic), Containers::StridedArrayView2D<typename Derived::Scalar>>::type arrayCast(const Eigen::DenseCoeffsBase<Derived, Eigen::DirectWriteAccessors>& from) {
```
mosra, line 411:
```suggestion
template<class Derived> inline typename std::enable_if<Eigen::internal::traits<Derived>::ColsAtCompileTime == Eigen::Dynamic && Eigen::internal::traits<Derived>::RowsAtCompileTime == Eigen::Dynamic, Containers::StridedArrayView2D<typename Derived::Scalar>>::type arrayCast(const Eigen::DenseCoeffsBase<Derived, Eigen::DirectWriteAccessors>& from) {
```

The ()s are not necessary, same above. Btw., both functions have the same argument list, how is the compiler not complaining? In my case I always had to move the SFINAE to the argument list (extra defaulted param, e.g.) to make it work.

src/Magnum/EigenIntegration/Integration.h
```diff
@@ -369,6 +370,51 @@ template<class To, std::size_t size, class T> inline To cast(const Math::Vector<
...
 
+
+/**
+@brief Convert Corrades StridedArrayView2D to Eigens Dynamic Matrix Type.
+*/
+template<class T> inline Eigen::Map<Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>, Eigen::Unaligned, Eigen::Stride<Eigen::Dynamic, Eigen::Dynamic>> arrayCast(const Containers::StridedArrayView2D<T>& from) {
+    using MatrixT = Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
+    using StrideT = Eigen::Stride<Eigen::Dynamic, Eigen::Dynamic>;
+    auto size = from.size();
+    auto stride = from.stride();
```
mosra, line 381:
```suggestion
    const Containers::StridedDimensions2D<std::size_t> size = from.size();
    const Containers::StridedDimensions2D<std::ptrdiff_t> stride = from.stride();
```

I'm a `const` and `auto` extremist, sorry :P

src/Magnum/EigenIntegration/Test/IntegrationTest.cpp
```diff
@@ -215,6 +231,78 @@ void IntegrationTest::matrixMatrix() {
...
+    for(int i = 0; i < 4*5; ++i) data[i] = float(i);
+    Containers::StridedArrayView2D<float> view1(data, {4, 5});
+    auto mapped = arrayCast(view1);
+    auto view2 = arrayCast(mapped);
+    for(int i = 0; i < 4; ++i) {
+        for(int j = 0; j < 5; ++j) {
+            CORRADE_COMPARE(view1[i][j], view2[i][j]);
+        }
+    }
+}
```
mosra, line 246: This could use an additional test case that 

1. uses a `flipped()`, `broadcasted()` and `transposed()` array on input to test all cases of a positive, zero and negative stride
2. converts to Eigen, verifies a few random picks (`a[2][3] == b[2][3]`)
3. converts back

and then verifies that `data()`, `size()` and `stride()` are the same as on input. In particular I'm not sure if Eigen even supports zero strides (I hope it supports negative strides at least).

src/Magnum/EigenIntegration/Test/IntegrationTest.cpp
```diff
@@ -215,6 +231,78 @@ void IntegrationTest::matrixMatrix() {
...
+void IntegrationTest::stridedArrayView() {
+    float data[4*5];
+    for(int i = 0; i < 4*5; ++i) data[i] = float(i);
+    Containers::StridedArrayView2D<float> view1(data, {4, 5});
+    auto mapped = arrayCast(view1);
+    auto view2 = arrayCast(mapped);
+    for(int i = 0; i < 4; ++i) {
+        for(int j = 0; j < 5; ++j) {
+            CORRADE_COMPARE(view1[i][j], view2[i][j]);
+        }
```
mosra, line 244:
```suggestion
        CORRADE_ITERATION(i);
        CORRADE_COMPARE_AS(view1[i], view2[i], TestSuite::Compare::Container);
```

You'll need `#include <Corrade/TestSuite/Compare/Container.h>` for this.

src/Magnum/EigenIntegration/Test/IntegrationTest.cpp
```diff
@@ -215,6 +231,78 @@ void IntegrationTest::matrixMatrix() {
...
+    Eigen::MatrixXf m = Eigen::MatrixXf::Random(4, 5);
+    auto row = m.row(1);
+    auto view1 = arrayCast(row);
+    auto mapped1 = arrayCast(view1);
+    /* need to transpose since arrayCast returns a column vector */
+    CORRADE_VERIFY(mapped1.isApprox(row.transpose()));
+
+    auto col = m.col(2);
+    auto view2 = arrayCast(col);
+    auto mapped2 = arrayCast(view2);
```
mosra, line 302: Can we do less `auto`? :sweat_smile: Same in all cases above.

src/Magnum/EigenIntegration/Integration.h
```diff
@@ -369,6 +370,51 @@ template<class To, std::size_t size, class T> inline To cast(const Math::Vector<
     return Math::Implementation::VectorConverter<size, T, To>::to(from);
 }
 
+
+/**
+@brief Convert Corrades StridedArrayView2D to Eigens Dynamic Matrix Type.
+*/
+template<class T> inline Eigen::Map<Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>, Eigen::Unaligned, Eigen::Stride<Eigen::Dynamic, Eigen::Dynamic>> arrayCast(const Containers::StridedArrayView2D<T>& from) {
+    using MatrixT = Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
```
mosra, line 378: Could be the `RowMajor` / `ColMajor` configurable via an optional second template parameter, maybe? Not sure what should be the default tho.

doc/snippets/EigenIntegration.cpp
```diff
@@ -64,6 +64,24 @@ Debug{} << Eigen::Array3i{1, 42, -3}; // prints   1
 static_cast<void>(d);
 }
 
+{
+/* [IntegrationStridedArrayView] */
+float data[15];
+Containers::StridedArrayView2D<float> view{data, {3,5}};
+
+/* To avoid a copy make sure to provide the correct type or use auto. */
+auto map = EigenIntegration::arrayCast(view);
```
mosra, line 73: Mention the type in a comment at least?

doc/snippets/EigenIntegration.cpp
```diff
@@ -64,6 +64,24 @@ Debug{} << Eigen::Array3i{1, 42, -3}; // prints   1
...
+
+/* Copy everything into Eigen Matrix. */
```
mosra, line 75:
```suggestion
/* Unlike the above, this copies everything */
```

... or something that makes the point clearer, I'm too tired to invent a better comment right now.

src/Magnum/EigenIntegration/Integration.h
Janos95: I think for two function templates to be considered equivalent, expressions involving template parameters in their return types have to be equivalent so the compiler has to evaluate the enable_if struct and one of them is always discarded. But good question why this is allowed since it is forbidden to overload on the return type...

src/Magnum/EigenIntegration/Test/IntegrationTest.cpp
Janos95: So I researched this a bit, and it is possible to incorporate negative strides by mapping those to Eigen::Reverse expressions I think.
Unfortunately I dont think it is possible to cast broadcasted ArrayViews, since Eigen handels braodcasting differently. E.g. to add a vector to a matrix one turns the matrix into a special kind of broadcast expression but the vector is not changed in any way 
```
Eigen::MatrixXf m; 
Eigen::VectorXf v;
m.rowwise() += v;
```
So if you agree I would just assert on zero strides.

src/Magnum/EigenIntegration/Integration.h
Janos95: I think having it RowMajor is consistent with how it is used in Magnum (i.e.. view2D[0] is usually used as a row, on the other hand the default in Eigen is ColumnMajor)  and one can always transpose the resulting Eigen expression to get a ColumnMajor view on the data if I am not mistaken?

src/Magnum/EigenIntegration/Test/IntegrationTest.cpp
Janos95: Ok actually the Reverse expression are not well suited since everything needs to be known at compile time. I found that negative strides are actually supported out of the box since relatively recently : https://eigen.tuxfamily.org/bz/show_bug.cgi?id=747 .

src/Magnum/EigenIntegration/Integration.h
mosra: I investigated and tried to support both `RowMajor` / `ColMajor` via a template parameter only to realize it doesn't really have any effect on a `Map` if you explicitly specify `Stride` as well: https://eigen.tuxfamily.org/dox/group__TutorialMapClass.html#TutorialMapTypes A `Map` can be then copied to both row- and col-major `Matrix` so unless I'm missing something it doesn't matter at all what `Matrix` type is used inside the `Map`.

So I removed the `RowMajor`, flipped order of `Stride` arguments and it resulted in various types being significantly shorter: 9379c6411a8ac95abc9bc0a7e5c2a8e95b0df4a7

src/Magnum/EigenIntegration/Test/IntegrationTest.cpp
mosra: Looking at the commit log, this seems to be done and fully supported, right? :)

mosra:
Finally merged as e8eceb3a27317d36f37af68fdd0fc593b83a51b3, thank you! :pray: (and sorry for the several month delay)

As I mentioned in [the comment above](https://github.com/mosra/magnum-integration/pull/74#discussion_r500491972), 9379c6411a8ac95abc9bc0a7e5c2a8e95b0df4a7 removes the `RowMajor` from the `Map` typedef -- can you quickly verify it makes sense? I'm not "Eigen native" so I'm not 100% sure this doesn't break anything. Thanks again!

PR #78, opened 2020-12-16
modules: fix linking error when build with latest imgui

linuxaged:
modules: fix `ImGui::TableSettingsInstallHandler(ImGuiContext*)` not found when building with latest version of imgui

ImGui's tables feature has been merged into master since [9874077](https://github.com/ocornut/imgui/commit/9874077fc0e364383ef997e3d4332172bfddc0b9), and the tables impl was moved into imgui_tables.cpp

### DIFF
--- a/modules/FindImGui.cmake
+++ b/modules/FindImGui.cmake
@@ -148,7 +148,7 @@ foreach(_component IN LISTS ImGui_FIND_COMPONENTS)
             set(ImGui_Sources_FOUND TRUE)
             set(ImGui_SOURCES )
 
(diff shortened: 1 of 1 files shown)


mosra:
Thank you! I'll see what needs to be done to make this work on the older versions as well (I suppose that's why all the CIs fail for this :)).

mosra:
Merged (a modified version for backwards compatibility) in 1f79ba71b9f0ece204dd9ba31b06871653c0d546, thank you! :+1:

PR #84, opened 2021-03-20
Add support for draw list user callbacks in ImGui integration

### LINKED ISSUE #83: ImGui integration does not support adding UserCallback to DrawList
The `ImDrawList` advanced API allows one to register arbitrary callbacks to change the OpenGL state during rendering. This can be used to manipulate shaders or, in my case, to temporarily change the `glBlendFunc` for the rendering of one particular widget. 

This needs support from the rendering engine. The Magnum integration is ignoring these, but this is how the standard OpenGL integration handles them:
https://github.com/ocornut/imgui/blob/6a161b878943241ddecbeee4ae27103dbd2d33d0/backends/imgui_impl_opengl2.cpp#L163

arximboldi:
Fixes #83

### DIFF
--- a/src/Magnum/ImGuiIntegration/Context.cpp
+++ b/src/Magnum/ImGuiIntegration/Context.cpp
@@ -318,23 +318,34 @@ void Context::drawFrame() {
 
         for(std::int_fast32_t c = 0; c < cmdList->CmdBuffer.Size; ++c) {
             const ImDrawCmd* pcmd = &cmdList->CmdBuffer[c];
+            if (pcmd->UserCallback)
+            {
+                // User callback, registered via ImDrawList::AddCallback()
+                // ImDrawCallback_ResetRenderState is a special callback value
+                // used by the user to request the renderer to reset render
+                // state. We do not have anything to do here though...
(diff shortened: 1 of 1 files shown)


pezcode:
Hey @arximboldi 👋 Any chance you could rebase on top of master and fix the code style to match the [Corrade coding style](https://doc.magnum.graphics/corrade/corrade-coding-style.html)?

Namely:
- `if ()` -> `if()`
- opening brackets on same line -> `if() {`

Ideally there would also be a test in `ContextGLTest.cpp` that checks that the callback is called correctly.

Let me know if you don't have the time or want me to do this, I don't mind taking over.

mosra:
This again slipped through my attention, sorry. At first I waited until we had the test scaffolding set up, but then just completely forgot to get back to this.

@pezcode if you're able to add the tests, that would be the most great. Coding-style-wise, don't bother, I can fix that during merge ;)

arximboldi:
I don't have time to work on this now so I would appreciate if someone can pick it up. Thank you a lot!

pezcode:
I opened https://github.com/mosra/magnum-integration/pull/98. I would've preferred to have just one PR, but this way we get CI tests running.

PR #85, opened 2021-07-16
ImGuiIntegration: Allow to include another imconfig

gergondet:
The integration defines `IMGUI_USER_CONFIG` to change the visibility of ImGui API. If one sets `IMGUI_USER_CONFIG` in their own project this will create a conflict.

By defining `MAGNUM_IMGUI_USER_CONFIG` one can include their own file along the ImGuiIntegration provided one.

Note: maybe it would be more inline with the rest of the library to have the macro named `MAGNUM_IMGUIINTEGRATION_USER_CONFIG`, I have no problem with changing the name if you feel this is better

Thanks for taking time to consider this and for the great work on the overall project!


mosra:
Hi, sorry for the delayed response, I have a constant notification overload :)

> maybe it would be more inline with the rest of the library to have the macro named `MAGNUM_IMGUIINTEGRATION_USER_CONFIG`

Yes, that sounds better. Can you change it?

Another thing, it might be a good idea to document this somewhere, because otherwise I fear the feature will get forgotten. Can you add a paragraph about the macro and why `IMGUI_USER_CONFIG` doesn't work to [this place](https://github.com/mosra/magnum-integration/blob/789f8132842536f077548bc4f2d2c3685e9f491e/doc/namespaces.dox#L392)? It would then [appear here](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1ImGuiIntegration.html) which I think is a good-enough location.

Thank you! :+1:

gergondet:
Hi  @mosra 

Thank you for the feedback and sorry for the delay on my side. I've taken you remarks into account. The macro has been renamed and I've added a small paragraph about the feature in the place you suggested.

Thanks!

mosra:
Sorry for the extremely delayed responses :see_no_evil: 

Thanks for the docs update :+1: Merged this as 1faf8d3e7c960ac57db80e90c48ce6aad19e86a1, for good measure I added a test as well -- was suspicious that the quotes in CMake could break on some systems, but apparently it works everywhere the CIs test for, so great!

Issue #87, opened 2022-01-11
Key Events have changed in ImGui


jvannugteren:
In the latest version of ImGui the key events enums have shifted by 512 preventing proper capture of keys such as delete and backspace. Now the function "AddKeyEvent" has to be used to do the translation.

I think the issue can be fixed by replacing the switch in src/Magnum/ImGuiIntegration/Context.hpp file to:
```
switch(event.key()) {
        /* LCOV_EXCL_START */
        case KeyEvent::Key::LeftShift:
        case KeyEvent::Key::RightShift:
            io.KeyShift = value;
            break;
        case KeyEvent::Key::LeftCtrl:
        case KeyEvent::Key::RightCtrl:
            io.KeyCtrl = value;
            break;
        case KeyEvent::Key::LeftAlt:
        case KeyEvent::Key::RightAlt:
            io.KeyAlt = value;
            break;
        case KeyEvent::Key::LeftSuper:
        case KeyEvent::Key::RightSuper:
            io.KeySuper = value;
            break;
        case KeyEvent::Key::Tab:
            io.AddKeyEvent(ImGuiKey_Tab, value);
            break;
        case KeyEvent::Key::Up:
            io.AddKeyEvent(ImGuiKey_UpArrow, value);
            break;
        case KeyEvent::Key::Down:
            io.AddKeyEvent(ImGuiKey_DownArrow, value);
            break;
        case KeyEvent::Key::Left:
            io.AddKeyEvent(ImGuiKey_LeftArrow, value);
            break;
        case KeyEvent::Key::Right:
            io.AddKeyEvent(ImGuiKey_RightArrow, value);
            break;
        case KeyEvent::Key::Home:
            io.AddKeyEvent(ImGuiKey_Home, value);
            break;
        case KeyEvent::Key::End:
            io.AddKeyEvent(ImGuiKey_End, value);
            break;
        case KeyEvent::Key::PageUp:
            io.AddKeyEvent(ImGuiKey_PageUp, value);
            break;
        case KeyEvent::Key::PageDown:
            io.AddKeyEvent(ImGuiKey_PageDown, value);
            break;
        case KeyEvent::Key::Enter:
        case KeyEvent::Key::NumEnter:
            io.AddKeyEvent(ImGuiKey_Enter, value);
            break;
        case KeyEvent::Key::Esc:
            io.AddKeyEvent(ImGuiKey_Escape, value);
            break;
        case KeyEvent::Key::Space:
[...]

PR #89, opened 2022-01-12
ImGuiIntegration: fix indexing into KeysDown

pezcode:
Somewhat bandaid-y fix for https://github.com/mosra/magnum-integration/issues/87. The tl;dr is that with imgui >1.86 `ImGuiKey_*` don't start at 0, so we need to subtract the first value to index `io.KeysDown` correctly.

This isn't a long-term fix since the whole `KeyMap` + `KeysDown` thing is deprecated and will be removed in the future.

### DIFF
--- a/src/Magnum/ImGuiIntegration/Context.cpp
+++ b/src/Magnum/ImGuiIntegration/Context.cpp
@@ -57,6 +57,8 @@ Context::Context(ImGuiContext& context, const Vector2& size, const Vector2i& win
     /* Ensure we use the context we're linked to */
     ImGui::SetCurrentContext(&context);
 
+    ImGuiIO &io = ImGui::GetIO();
+
     /* The KeyMap is meant to be for mapping from engine-native zero-based
        enums to ImGui enums in order to avoid complex switch-case and allow
        users to use native enums with ImGui input APIs. However Magnum only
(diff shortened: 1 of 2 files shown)


jvannugteren:
This fix works fine for my code.

ocornut:
I'm particularly interested in this as our change https://github.com/ocornut/imgui/issues/4858 was designed to completely backward compatible, if any code change is required by a backend we have an issue on our side.

Seeing the PR I realize how this specific use would indeed cause a problem:

```
io.KeyMap[ImGuiKey_Tab] = ImGuiKey_Tab;
```

I will add an assert on dear imgui side if a legacy KeyMap[] entry is >=512 and document this specific use a breaking change on our side.

ocornut:
> I will add an assert on dear imgui side if a legacy KeyMap[] entry is >=512 and document this specific use a breaking change on our side.

I tested this and there was already a triggering assert (good), still going to update documentation to mark this use as breaking.

mosra:
Hmm, so if I understand correctly, this is obsoleted by #93 (which is still a draft), right? Thus nothing to do for me with this one.

pezcode:
> Hmm, so if I understand correctly, this is obsoleted by #93 (which is still a draft), right? Thus nothing to do for me with this one.

Correct, I'll close this to avoid any confusion 🙇‍♂️

Issue #82, opened 2021-03-16
ImGui integration does not support drawing lots of elements (ImGuiBackendFlags_RendererHasVtxOffset)


arximboldi:
Hi!

When drawing meshes with more than 64K elements with ImGui the backend needs to support `ImGuiBackendFlags_RendererHasVtxOffset`

Looking at it, it should not be very difficult for the Magnum integration to do this. One can check their official OpenGL3 renderer for reference:
https://github.com/ocornut/imgui/blob/6a161b878943241ddecbeee4ae27103dbd2d33d0/backends/imgui_impl_opengl3.cpp

Cheers!

mosra:
Hi, 

I was aware of this feature and it was on my TODO list for quite some time, but I treated it more like a perf optimization than a blocker (IIRC it can reduce some work this way), which is why it's not there yet :) Coincidentally, Magnum recently got support for vertex offset for GLES and WebGL (for some reason it's an extreme problem to specify vertex offset on those platforms), so this feature could be added not just for desktop GL.

Unfortunately at this point I'm rather busy with other things so I can't promise any ETA when this gets done. 

(Or maybe @Squareys could have an use for this as well? Haha.)

arximboldi:
No worries! I worked-around this so far by using `ImDrawIdx=unsigned int`. I agree this can be considered an optimization rather than a bug (it's only mildly annoying because one has to figure out what is going on when one starts experiencing data truncation in ImGui).

In any case, thanks for the super-quick response and all the great work in this library. Magnum is a very well engineered library! :heart:

mosra:
Ah, if that works that's great.

Btw., can you share what you're using Magnum for? :) I'm wondering what use case involves drawing thousands of GUI elements.

arximboldi:
Sadly I can't share but I can tell you it's a cool project in the domain of movie-making :)

The ImGui limit was hit though while using this library to plot dense graphs: https://github.com/epezent/implot

mosra:
Fixed with #90.

arximboldi:
Thank you so much @mosra :clap:

PR #90, opened 2022-01-12
ImGuiIntegration: add tests, support ImDrawCmd::VtxOffset

pezcode:
Closes https://github.com/mosra/magnum-integration/issues/82.

I'm not 100% sure if the version/extension checks are correct this way. Info was taken from https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1Mesh.html#aaa49afa2c76cc12b8402418a6e31923f.

~~This also needs some more testing on GLES/WebGL, hence a draft PR.~~


pezcode:
I added a small test window to visually check for correct handling of `VtxOffset`, taken from [here](https://github.com/ocornut/imgui/issues/2591#issuecomment-496954460).

If all goes well, it should look something like this with no obvious glitches:

chrome_2022-01-14_22-00-33

(screenshot taken on Chrome 97, with *WebGL Draft Extensions* enabled)

pezcode:
Forgot to mention: this unconditionally requires imgui 1.71. Since that version is considered obsolete (ie. no official backward compatibility in the API), I'd say that's acceptable. If you'd prefer to support older versions, we can `#ifdef` around this as well.

src/Magnum/ImGuiIntegration/Context.cpp
```diff
@@ -324,6 +342,8 @@ void Context::drawFrame() {
                 {pcmd->ClipRect.z, fbSize.y() - pcmd->ClipRect.y}}
                     .scaled(_supersamplingRatio)});
 
+            /* Only > 0 if ImGuiBackendFlags_RendererHasVtxOffset is set */
+            _mesh.setBaseVertex(pcmd->VtxOffset);
             _mesh.setCount(pcmd->ElemCount);
             _mesh.setIndexBuffer(_indexBuffer, pcmd->IdxOffset*sizeof(ImDrawIdx),
```
mosra, line 348: IIRC, the same effect could be achieved by having a 32-bit ImDrawIdx, right? As a fallback solution, especially since the `WEBGL_draw_instanced_base_vertex_base_instance` extension is not that widely available as far as I know.

As far as I know, it's not runtime-switchable, has to be defined in user code before including anything from imgui and thus we can't control it ourselves. I think it would make sense to have that documented at least -- for example [before the Multiple contexts section](https://github.com/mosra/magnum-integration/blob/af8b343b3466e92a7cbcdd4a53757150ea754273/src/Magnum/ImGuiIntegration/Context.h#L253).

Actually, it would be good to have it documented together with the extension requirements ("use 32-bit indices if you can't ensure the extensions are present, especially on WebGL"), and maybe mentioning that [ImPlot is a widget often pushing the limits](https://github.com/mosra/magnum-integration/issues/82#issuecomment-800179085). Thanks!

pezcode:
I added integration tests for `renderFrame()` to make sure it correctly handles scissor rect, texture IDs, index offset and vertex offset (if supported). Hopefully makes your Vulkan porting easier in the future, too 🌋 The visual test got removed because it's now unnecessary.

All the scaffolding was taken from the Magnum Shader tests. Some special-casing (iOS?) may not be needed, I wasn't too sure about the interactions there.

pezcode:
Not sure what's happening to Github, it just completely skipped running CI 💩

pezcode:
The linux-gles tests seem to be the only ones that run ctest without `-E GLTest`. Can we enable the GL tests in other CIs without too much breaking? I assume they're disabled for a good reason, but running the ImGui `drawFrame()` tests on a few different platforms would be useful.

src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp
```diff
@@ -176,13 +210,46 @@ ContextGLTest::ContextGLTest() {
...
+        /** @todo Fix this once I persuade CMake to run XCTest tests properly */
+        && std::getenv("SIMULATOR_UDID")
+        #endif
+    ) {
+        _testDir = Utility::Directory::path(Utility::Directory::executableLocation());
+    } else
+    #endif
+    {
+        _testDir = SHADERS_TEST_DIR;
+    }
```
mosra, line 252: Same here, let's just ignore iOS for now.

src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp
```diff
@@ -148,6 +166,22 @@ struct ContextGLTest: GL::OpenGLTester {
...
+
+    void draw();
+    void drawTexture();
+    void drawScissor();
+    void drawVertexOffset();
+    void drawIndexOffset();
+
+    private:
+        PluginManager::Manager<Trade::AbstractImporter> _manager{"nonexistent"};
+        std::string _testDir;
```
mosra, line 181:
```suggestion
        PluginManager::Manager<Trade::AbstractImporter> _manager;
```

And the `_testDir` can be just ~~`SHADERS_TEST_DIR`~~ `IMGUIINTEGRATION_TEST_DIR` now.

src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp
```diff
@@ -176,13 +210,46 @@ ContextGLTest::ContextGLTest() {
...
     GL::Renderer::enable(GL::Renderer::Feature::ScissorTest);
+
+    /* Load the plugins directly from the build tree. Otherwise they're either
+       static and already loaded or not present in the build tree */
+    #ifdef ANYIMAGEIMPORTER_PLUGIN_FILENAME
+    CORRADE_INTERNAL_ASSERT_OUTPUT(_manager.load(ANYIMAGEIMPORTER_PLUGIN_FILENAME) & PluginManager::LoadState::Loaded);
+    #endif
+    #ifdef TGAIMPORTER_PLUGIN_FILENAME
+    CORRADE_INTERNAL_ASSERT_OUTPUT(_manager.load(TGAIMPORTER_PLUGIN_FILENAME) & PluginManager::LoadState::Loaded);
+    #endif
```
mosra, line 238: No need for this anymore.

src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp
```diff
@@ -742,6 +809,282 @@ void ContextGLTest::multipleContexts() {
...
+    drawList->AddTriangleFilled({0.0f, 0.0f}, {size.x*0.5f, 0.0f}, {size.x*0.25f, size.y*0.5f},
+        IM_COL32(128, 255, 128, 255));
+
+    c.drawFrame();
+
+    MAGNUM_VERIFY_NO_GL_ERROR();
+
+    if(!(_manager.loadState("AnyImageImporter") & PluginManager::LoadState::Loaded) ||
+       !(_manager.loadState("TgaImporter") & PluginManager::LoadState::Loaded))
+        CORRADE_SKIP("AnyImageImporter / TgaImporter plugins not found.");
```
mosra, line 872: Consistently with how it's handled e.g. in case of `KtxImporter` in `AnyImageImporter` tests as of mosra/magnum@578baf3168b50da73d692d48ffa878c19e313baa -- using `load()` will catch not only absence of the plugin, but also a case when it doesn't link because of an ABI change:

```suggestion
    /* Catch also ABI and interface mismatch errors */
    if(!(manager.load("AnyImageImporter") & PluginManager::LoadState::Loaded) ||
       !(manager.load("PngImporter") & PluginManager::LoadState::Loaded))
        CORRADE_SKIP("AnyImageImporter / PngImporter plugin can't be loaded.");
```

Similarly in the other case below.

(I already switched to PNGs here, yes.)

src/Magnum/ImGuiIntegration/Context.h
```diff
@@ -251,6 +251,35 @@ If you don't do that, the fonts stay at the original scale, not matching the
 new UI scaling anymore. If you didn't supply any custom font, the function will
 reconfigure the builtin font automatically.
 
+@section ImGuiIntegration-Context-large-meshes Large meshes
+
+Complex user interfaces or widgets like [ImPlot](https://github.com/epezent/implot)
+may end up creating large meshes with more than 64,000 vertices. Because ImGui
```
mosra, line 257:
```suggestion
may end up creating large meshes with more than 65k vertices. Because ImGui
```

src/Magnum/ImGuiIntegration/Context.h
```diff
@@ -251,6 +251,35 @@ If you don't do that, the fonts stay at the original scale, not matching the
...
+defaults to 16-bit index buffers this can lead to asserts or visual errors.
+
+If the underlying GL context supports @ref GL::Mesh::setBaseVertex() "setting the base vertex for indexed meshes"
+the rendering backend sets the @cpp ImGuiBackendFlags_RendererHasVtxOffset @ce
+flag. This lets ImGui know the backend can handle per-draw vertex offsets,
+allowing meshes with a little over 2 million vertices. Support for that
```
mosra, line 263: Where does 2 million come from? Is this some ImGui limitation? IIUC, the vertex offset is a signed int, so 2 *billion* should be theoretically possible ... but I'd just not mention this at all, doesn't seem that important.

```suggestion
removing the 65k limitation altogether. Support for that
```

src/Magnum/ImGuiIntegration/Context.h
```diff
@@ -251,6 +251,35 @@ If you don't do that, the fonts stay at the original scale, not matching the
...
+requires one of the following:
+
+-   OpenGL 3.2 and higher or extension `ARB_draw_elements_base_vertex`
+-   OpenGL ES 3.2 or extension `OES_draw_elements_base_vertex` or
+    `EXT_draw_elements_base_vertex`
+-   WebGL 2.0 and extension `WEBGL_draw_instanced_base_vertex_base_instance`
```
mosra, line 269: There's a few custom Doxygen commands that provide links to extensions and cross-link this page with an index of all features requiring a particular version/extension:

```suggestion
@requires_gl32 Extension @gl_extension{ARB,draw_elements_base_vertex}
@requires_gles32 Extension @gl_extension{OES,draw_elements_base_vertex} or
    @gl_extension{EXT,draw_elements_base_vertex}
@requires_webgl_extension WebGL 2.0 and extension
    @webgl_extension{WEBGL,draw_instanced_base_vertex_base_instance}
```

mosra (commented):
> Can we enable the GL tests in other CIs without too much breaking?

The GL tests run against SwiftShader, which is the only software GL implementation I was able to get working there, and it's GLES-only. I could enable it on macOS as well as Magnum builds got that recently already, on Windows I don't have any existing SwiftShader build to copy from yet, so there it will take some more effort.

After I manage to tag a release I plan to drop support for Ubuntu 16.04, and then I could try adding a Mesa llvmpipe build, to have an ability to test the full desktop GL. But that's still quite far ahead.

pezcode:
Thanks for the detailed review ☺️

I think I fixed everything, except for replacing `TgaImporter` with `PngImporter`. The effort of adding and building `magnum-plugins` on all CIs just for the `ImGuiIntegration` GL tests didn't seem worth it considering the test images are 2.5KB combined. But if you really have a strong preference for PNGs, I can add it.

src/Magnum/ImGuiIntegration/Context.h
pezcode: Math is not my strong suit 🤡

src/Magnum/ImGuiIntegration/Test/CMakeLists.txt
```diff
@@ -41,11 +41,79 @@ set_target_properties(
     PROPERTIES FOLDER "Magnum/ImGuiIntegration/Test")
 
 if(BUILD_GL_TESTS)
-    corrade_add_test(ImGuiContextGLTest ContextGLTest.cpp
-        LIBRARIES MagnumImGuiIntegration Magnum::OpenGLTester)
+    # Otherwise CMake complains that Corrade::PluginManager is not found, wtf
```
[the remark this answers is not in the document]
pezcode: See my comment below about PNGs, but I'd be happy to revisit this, too

mosra:
Since it's a UI library, I anticipate the test image corpus to grow over time, especially if/when we switch to our own text rendering, or to Magnum's own theme, for example. With small images TGAs are acceptable, but when they grow larger it stops being practical due to the huge file sizes. (Not to mention the lack of software support makes dealing with them kinda annoying, especially here in the browser.)

> adding and building magnum-plugins on all CIs

Just on the ES build, in `unix-desktop-gles.sh`, where the GL tests are run ;) If you pick StbImageImporter (and not the full PngImporter with the external dependencies), apart from the time it takes to clone the repo & run cmake, the build time would be comparable to build time of the TgaImporter. Nothing else from the plugins repo will get built, I made it so exactly this use case is as efficient as it can be.

You can reuse the relevant piece from `unix-desktop.sh`, just drop the DART-related stuff: https://github.com/mosra/magnum-integration/blob/26e7ca90aa0b4b4c406c8384f18f57c0c474098a/package/ci/unix-desktop.sh#L51-L64

mosra (approved):
:tada:

pezcode:
Just realized the `TgaImporter` is still being built, as well as `AnyImageImporter` on CIs that don't run the GL tests. If you're not in a rush, I'll fix it sometime this week. 🙇

mosra:
Merged as 26e7ca90aa0b4b4c406c8384f18f57c0c474098a...7931348bc797a7f8794452e00d9a0aac3faebc0c, thanks for the reminder -- I changed the commits to enable the relevant plugins only for the ES CI builds.

As of b29aeeed8e083f3f25f9c1921f1f259a3ee50c81 there's also a GLES3 macOS build, I'll eventually add Windows variants once Magnum has them.

PR #92, opened 2022-01-19
ImGuiIntegration: support ImGuiMouseCursor_NotAllowed

pezcode:
This sets the No cursor (_"slashed circle or crossbones"_ = 🚫) for `ImGuiMouseCursor_NotAllowed` (available in imgui 1.75+) when using an Application class that has `Cursor::No`.


mosra:
Thank you! Merged as 395631f0497d10a711a7bf209817fd70e324c79a, this one was easy.

PR #93, opened 2022-01-19
ImguiIntegration: support for imgui IO event API

pezcode:
This adds support for the [new IO handling](https://github.com/ocornut/imgui/issues/4858) about to land in imgui 1.87. The main advantages:
- Full range of keys that roughly matches the `Application` classes' `KeyEvent::Key`, meaning you can do key handling with imgui, independent of the `Application` class
- Input events are trickled, e.g. mouse button down followed by mouse button up in the same frame is now spaced out over two frames. This makes the manual workaround for mouse buttons obsolete and fixes the same issue for keys.

A few notes:
- There is full backward compatibility with older imgui versions, using the legacy IO interface
- These changes make https://github.com/mosra/magnum-integration/pull/89 obsolete
- [x] Depends on https://github.com/mosra/magnum/pull/547 so all `Application` classes support the entire range of `ImGuiKey`

Todo:
- [x] Test with 1.87 final and update the API version used to check for the new IO API


src/Magnum/ImGuiIntegration/Context.hpp
```diff
@@ -40,93 +40,161 @@
 #include "Magnum/ImGuiIntegration/Integration.h"
 #include "Magnum/ImGuiIntegration/Context.h"
 
+#define HAS_NEW_IMGUI_IO (IMGUI_VERSION_NUM >= 18615)
```
mosra, line 43: If you prefix the macro and hide it from Doxygen you can leave it defined, without having to `#undef` it at the end, and without having to define it again in the `*.cpp` file:

```suggestion
#if !defined(DOXYGEN_GENERATING_OUTPUT) && IMGUI_VERSION_NUM >= 18615
#define MAGNUM_IMGUIINTEGRATION_HAS_NEW_IMGUI_IO
#endif
```

I also changed it to be defined only if it's true, mainly for consistency with other defines out there.

src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp
```diff
@@ -36,18 +36,26 @@
...
-enum class Modifier: Int {
-    Shift, Ctrl, Alt, Super
+struct InputEvent {
+    enum class Modifier: Int {
+        Shift = 1 << 0,
+        Ctrl = 1 << 1,
+        Alt = 1 << 2,
+        Super = 1 << 3
+    };
+    typedef Containers::EnumSet<Modifier> Modifiers;
```
mosra, line 46: You could move the `_modifiers` and `modifiers()` here as well, to deduplicate the code a bit.

mosra (commented):
Thanks! I'll need a bit more time to wrap my head around all the changes, so this is just a partial review.

src/Magnum/ImGuiIntegration/Context.hpp
pezcode: Since `Context.cpp` includes `Context.h` before `imgui.h` I still need to duplicate this (otherwise `IMGUI_VERSION_NUM` isn't defined yet when we need it). Should I just move the into a `configure.h`?

mosra: You mean `configure.h.cmake`, right? Yes, that's a good place actually. And since it's ignored by Doxygen you don't even need the `#if !defined(DOXYGEN_GENERATING_OUTPUT)` there.

pezcode: There aren't any CMake variables we need so it could just be `.h` in the source folder. But I can go through CMake for consistency.

mosra: Um, there already *is* `configure.h.cmake`, needed for distinguishing static/dynamic build ;)

mosra:
Merged as c60149c818bbdf61d97f30706f2018d968930718, thanks a lot!

PR #96, opened 2022-06-17
ImGuiIntegration: Add conversion between ImStrv and StringView

hugoam:
Hello !
This one is kind of self-explanatory :)
Do you think we could merge something in this spirit?

### DIFF
--- a/src/Magnum/ImGuiIntegration/Integration.h
+++ b/src/Magnum/ImGuiIntegration/Integration.h
@@ -57,11 +57,27 @@ Example usage:
 #include "Magnum/ImGuiIntegration/visibility.h" /* defines IMGUI_API */
 
 #include <imgui.h>
+#include <Corrade/Containers/StringView.h>
 #include <Magnum/Types.h>
 #include <Magnum/Math/Vector.h>
 
-/* Don't list (useless) Magnum and Math namespaces without anything else */
+/* Don't list (useless) Corrade and Magnum and Math namespaces without anything else */
 #ifndef DOXYGEN_GENERATING_OUTPUT
+
+/* Currently present only in the features/string_view branch of ImGui */
+#ifdef IMGUI_HAS_IMSTR
+namespace Corrade { namespace Containers { namespace Implementation {
+
+template<> struct StringViewConverter<const char, ImStrv> {
+    static StringView from(const ImStrv& other) {
+        return StringView{other.Begin, other.length()};
+    }
+    static ImStrv to(StringView other) { return ImStrv{other.begin(), other.end()}; }
+};
+
+}}}
+#endif
+
 namespace Magnum { namespace Math { namespace Implementation {
 
 /* ImVec2 */
--- a/src/Magnum/ImGuiIntegration/Test/IntegrationTest.cpp
+++ b/src/Magnum/ImGuiIntegration/Test/IntegrationTest.cpp
@@ -35,6 +35,9 @@ namespace Magnum { namespace ImGuiIntegration { namespace Test { namespace {
 struct IntegrationTest: TestSuite::Tester {
     explicit IntegrationTest();
 
+#ifdef IMGUI_HAS_IMSTR
+    void stringView();
+#endif
     void vector2();
     void vector4();
     void color();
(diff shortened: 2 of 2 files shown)


mosra:
Yes, absolutely! :) 

Couldn't find this in the main imgui branch, so I suppose it's originating from this PR, right? https://github.com/ocornut/imgui/pull/3038 Can you add a bit of documentation, in a similar spirit to the other types listed above? Especially mentioning the PR/branch, and the `IMGUI_HAS_IMSTR` macro. And a test case as well, into `Test/IntegrationTest.cpp`. Thanks!

src/Magnum/ImGuiIntegration/Integration.h
```diff
@@ -57,9 +57,25 @@ Example usage:
...
+template <> struct StringViewConverter<const char, ImStrv> {
+    static StringView from(const ImStrv& other) {
+        return StringView{other.Begin, other.length()};
+    }
+    static ImStrv to(StringView other) { return ImStrv{other.begin(), other.end()}; }
+};
+
+}}}
+#endif
+#endif
```
mosra, line 77: Instead of closing `#ifndef DOXYGEN_GENERATING_OUTPUT` here, you could just move the below `#ifndef` above, including the comment.

src/Magnum/ImGuiIntegration/Integration.h
```diff
@@ -57,9 +57,25 @@ Example usage:
...
 #include <imgui.h>
+#include <Corrade/Containers/StringView.h>
 #include <Magnum/Types.h>
 #include <Magnum/Math/Vector.h>
 
+#ifndef DOXYGEN_GENERATING_OUTPUT
+#ifdef IMGUI_HAS_IMSTR
+namespace Corrade { namespace Containers { namespace Implementation {
+
+template <> struct StringViewConverter<const char, ImStrv> {
```
mosra, line 68:
```suggestion
template<> struct StringViewConverter<const char, ImStrv> {
```

Sorry, being picky :P

src/Magnum/ImGuiIntegration/Integration.h
```diff
@@ -57,9 +57,25 @@ Example usage:
 #include "Magnum/ImGuiIntegration/visibility.h" /* defines IMGUI_API */
 
 #include <imgui.h>
+#include <Corrade/Containers/StringView.h>
 #include <Magnum/Types.h>
 #include <Magnum/Math/Vector.h>
 
+#ifndef DOXYGEN_GENERATING_OUTPUT
+#ifdef IMGUI_HAS_IMSTR
```
mosra, line 65:
```suggestion
/* Currently present only in the features/string_view branch of ImGui */
#ifdef IMGUI_HAS_IMSTR
```

... or something like that, to make it clear why the ifdef is there.

mosra:
Apologies for cancelling the CI, I have *very little* free CircleCI credits left for June and saving them for emergency reasons :fire: 

If you can wait with pushing until July 1st, that'd be best. Thanks :sweat_smile:

src/Magnum/ImGuiIntegration/Test/IntegrationTest.cpp
```diff
@@ -35,19 +35,40 @@ namespace Magnum { namespace ImGuiIntegration { namespace Test { namespace {
...
+#ifdef IMGUI_HAS_IMSTR
+void IntegrationTest::stringView() {
+    using Containers::StringView;
+    StringView stringView("Hello World!");
+    ImStrv imStrv{"Hello World!"};
+
+    CORRADE_COMPARE(StringView(imStrv), stringView);
+
+    ImStrv c(stringView);
+    CORRADE_COMPARE(c, stringView);
```
mosra, line 68: Hm, as far as I can see, this would convert `c` to a StringView and then use Corrade's own equality comparison. Which I don't think is wanted, since it is no different from the other comparison above.

Instead, isn't there something like `ImStrCmp` that could be used here?

src/Magnum/ImGuiIntegration/Integration.h
```diff
@@ -57,11 +57,27 @@ Example usage:
 #include "Magnum/ImGuiIntegration/visibility.h" /* defines IMGUI_API */
```
mosra, line 57: Update the docs above to mention the string conversion, please ;)

Also feel free to add yourself to the copyright header in the license block above the file.

src/Magnum/ImGuiIntegration/Integration.h
```diff
@@ -57,11 +57,27 @@ Example usage:
...
 
 #include <imgui.h>
+#include <Corrade/Containers/StringView.h>
```
mosra, line 60: This would ideally be also wrapped in `#ifdef IMGUI_HAS_IMSTR`. 

Alternatively I was thinking of deinlining the conversion into a new `Integration.cpp` file and having just a forward declaration here, but that's maybe taking it a bit too far. The `StringView` header is not that huge.

mosra (commented):
Okay, CIs are capable of running again :)

Be sure to rebase on master, as there were some breaking changes in Magnum that would otherwise cause the CI to fail to build.

mosra:
@hugoam ping ;)

hugoam:
Sorry, didn't have much time to look at this again and then I went on a vacation. I'll try to have a look soonish, I promise

mosra:
Superseded by #122 and merged as 724732604c482ef58fec4553b4121d3009c42024...5b006b625044cf4d01efdb120b02479f9c9959bb. Thank you!

PR #98, opened 2022-07-28
ImGuiIntegration: add support for draw list user callbacks

pezcode:
Continuation of https://github.com/mosra/magnum-integration/pull/84 with tests

(*and an unrelated fix for a failing test in ImGui 1.88* :eyes:)

Closes https://github.com/mosra/magnum-integration/pull/84.

### DIFF
--- a/src/Magnum/ImGuiIntegration/Context.cpp
+++ b/src/Magnum/ImGuiIntegration/Context.cpp
@@ -334,6 +334,16 @@ void Context::drawFrame() {
         for(std::int_fast32_t c = 0; c < cmdList->CmdBuffer.Size; ++c) {
             const ImDrawCmd* pcmd = &cmdList->CmdBuffer[c];
 
+            if(pcmd->UserCallback) {
+                /* User callback, registered via ImDrawList::AddCallback().
+                   ImDrawCallback_ResetRenderState is a special callback value
+                   used by the user to request the renderer to reset render
+                   state. We do not have anything to do here though. */
+                if(pcmd->UserCallback != ImDrawCallback_ResetRenderState)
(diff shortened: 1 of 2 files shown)


mosra:
Amazing, I got nothing to complain about. Feels strange :D

Tangential: seeing the workaround for ImGui < 1.80, would it make sense to again bump the ImGui version that's used on the CI? The 1.72 is probably quite ancient nowadays. I think we discussed this before, but don't remember what the conclusion was.

pezcode:
Thanks for the speedy review 😌

> would it make sense to again bump the ImGui version that's used on the CI? The 1.72 is probably quite ancient nowadays. I think we discussed this before, but don't remember what the conclusion was.

A bump would make sense, yeah. If you don't want to spend CI minutes on two versions, the best one would be what the majority of people are likely using. 1.83 is about a year old, but might be too fresh for your taste 😄 
The good part about changes in this part of the repo is that you can change ImGui versions and run tests in a matter of 4 seconds. So keeping backwards compatibility just requires some minor due diligence on new PRs.

mosra:
Wasn't there some policy that if an ImGui version gets older than 2 years, it becomes "obsolete"? That's what I would aim for, together with regularly cleaning up code needed only for older versions... like we do with Assimp, for example.

Looking through the releases, 1.78 is from August 2020, so 1.77 would be the oldest one, 1.80 is still some months away :) There are a few projects, demos and examples that don't use ImGui that heavily so they have whatever version was current at the time and don't upgrade ever again. I don't want to break these "just because" but I also don't want to have extra maintenance overhead keeping ancient versions working. I'll bump to 1.77 then, and next time there are additions to this library, we can bump & cleanup again.

PR #100, opened 2022-10-21
ImGuiIntegration: don't use obsolete enum aliases

sthalik:
This is useful to me because clangd chokes on the oversized ImGui header and reducing its size through `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` helps a bit.


pezcode:
This would break for anyone using 1.87 and 1.88, since those enums only got renamed in 1.89 WIP: https://github.com/ocornut/imgui/commit/fd408c97904fbf9f3b4c786318e8389de11bc681

Best solution I can think of is to use the new enums optionally behind `#if IMGUI_VERSION_NUM >= 18823`. Might be a bit messy, but supporting `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` is indeed valuable.

mosra:
Thanks!

PR #102, opened 2023-04-23
ImGui: Add support for ImGuiBackendFlags_HasSetMousePos and cleaner drawFrame() code

Auburn:
`io.WantSetMousePos` will never be set unless requested by the user in the config like so:
```
ImGui::GetIO().ConfigFlags |= ImGuiConfigFlags_NavEnableSetMousePos;
ImGui::GetIO().ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard;
```

Minor tidy in DrawFrame(), the values in the ImDrawData are set based on the ImGui::GetIO() values so they will be the same

### DIFF
--- a/src/Magnum/ImGuiIntegration/Context.cpp
+++ b/src/Magnum/ImGuiIntegration/Context.cpp
@@ -85,6 +85,8 @@ Context::Context(ImGuiContext& context, const Vector2& size, const Vector2i& win
 
     /* Tell ImGui that changing mouse cursors is supported */
     io.BackendFlags |= ImGuiBackendFlags_HasMouseCursors;
+    /* We can honor io.WantSetMousePos requests */
+    io.BackendFlags |= ImGuiBackendFlags_HasSetMousePos;
 
     /* Check if we can support base vertex > 0 in draw commands */
     #if !(defined(MAGNUM_TARGET_WEBGL) && defined(MAGNUM_TARGET_GLES2))
--- a/src/Magnum/ImGuiIntegration/Context.cpp
+++ b/src/Magnum/ImGuiIntegration/Context.cpp
@@ -85,9 +85,7 @@ Context::Context(ImGuiContext& context, const Vector2& size, const Vector2i& win
 
     /* Tell ImGui that changing mouse cursors is supported */
     io.BackendFlags |= ImGuiBackendFlags_HasMouseCursors;
-    /* We can honor io.WantSetMousePos requests */
-    io.BackendFlags |= ImGuiBackendFlags_HasSetMousePos;
-
+    
     /* Check if we can support base vertex > 0 in draw commands */
     #if !(defined(MAGNUM_TARGET_WEBGL) && defined(MAGNUM_TARGET_GLES2))
     if(
--- a/src/Magnum/ImGuiIntegration/Context.cpp
+++ b/src/Magnum/ImGuiIntegration/Context.cpp
@@ -312,17 +312,17 @@ void Context::drawFrame() {
 
     ImGui::Render();
 
-    ImGuiIO& io = ImGui::GetIO();
-    const Vector2 fbSize = Vector2{io.DisplaySize}*Vector2{io.DisplayFramebufferScale};
-    if(!fbSize.product()) return;
-
     ImDrawData* drawData = ImGui::GetDrawData();
     CORRADE_INTERNAL_ASSERT(drawData); /* This is always valid after Render() */
-    drawData->ScaleClipRects(io.DisplayFramebufferScale);
+
+    const Vector2 fbSize = Vector2{drawData->DisplaySize}*Vector2{drawData->FramebufferScale};
+    if(!fbSize.product()) return;
+
+    drawData->ScaleClipRects(drawData->FramebufferScale);
 
     const Matrix3 projection =
         Matrix3::translation({-1.0f, 1.0f})*
-        Matrix3::scaling({2.0f/Vector2(io.DisplaySize)})*
+        Matrix3::scaling({2.0f/Vector2(drawData->DisplaySize)})*
         Matrix3::scaling({1.0f, -1.0f});
     _shader.setTransformationProjectionMatrix(projection);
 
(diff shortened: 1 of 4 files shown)


Auburn:
I added the mouse pos update into `Context::updateApplicationCursor` because it was convenient and had access to the application. But it could go in it's own function.

Auburn:
Looks like there is no way to warpCursor outside of SDL/GLFW applications, what is the best way to detect these are active?

mosra:
There's a test that mocks up the `Application` class in order to be able to verify the behavior in an automated headless setting, so it'd be about adding a `warpCursor()` implementation around here:

https://github.com/mosra/magnum-integration/blob/1a66b05bd7db0a5484366054ddc678bebf79921e/src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp#L117

and then ideally also adding a new test case verifying that this API gets correctly called when it should, somewhere near to the other mouse-related tests. Hopefully you can figure that out from the existing test code, feel free to ask if not.

The nastier problem is the fallback for apps without `warpCursor()` as you mentioned. I think it could be implemented via some SFINAE trick, unfortunately there isn't anything remotely similar to get an inspiration from, only the cursor-related stuff. But I think something like this could work:

```cpp
namespace Implementation {

template<class Application> void callWarpCursor(Application&, ...) {}
template<class Application, class = decltype(&Application::warpCursor)> void callWarpCursor(Application& application, const Vector2i& position) {
    application.warpCursor(position);
}

}
```

and then doing `Implementation::callWarpCursor(application, ...)` instead of `application.warpCursor()`. You'll also need some similar trick for setting the `ImGuiBackendFlags_HasSetMousePos` flag.

No guarantee if this works on all compilers tho :D

src/Magnum/ImGuiIntegration/Context.cpp
```diff
@@ -312,17 +314,17 @@ void Context::drawFrame() {
...
 
-    ImGuiIO& io = ImGui::GetIO();
-    const Vector2 fbSize = Vector2{io.DisplaySize}*Vector2{io.DisplayFramebufferScale};
-    if(!fbSize.product()) return;
-
     ImDrawData* drawData = ImGui::GetDrawData();
     CORRADE_INTERNAL_ASSERT(drawData); /* This is always valid after Render() */
-    drawData->ScaleClipRects(io.DisplayFramebufferScale);
+
+    const Vector2 fbSize = Vector2{drawData->DisplaySize}*Vector2{drawData->FramebufferScale};
```
mosra, line 320: Hm, I can't tell if this is a good or a bad change as my knowledge of ImGui is rather bad :sweat_smile: How does `drawData->FramebufferScale` and `io.DisplayFramebufferScale` relate?

Cc: @pezcode, you're responsible for most of the code here I think, and I bet you know better than me :)

Auburn: From the ImGui comments:
```
ImVec2          DisplayPos;             // Top-left position of the viewport to render (== top-left of the orthogonal projection matrix to use) (== GetMainViewport()->Pos for the main viewport, == (0.0) in most single-viewport applications)
ImVec2          DisplaySize;            // Size of the viewport to render (== GetMainViewport()->Size for the main viewport, == io.DisplaySize in most single-viewport applications)
ImVec2          FramebufferScale;       // Amount of pixels for each unit of DisplaySize. Based on io.DisplayFramebufferScale. Generally (1,1) on normal display, (2,2) on OSX with Retina display.
```

Auburn: It's mainly to aid with multi-window rendering 😉

Auburn:
>You'll also need some similar trick for setting the `ImGuiBackendFlags_HasSetMousePos` flag.

The problem with this is there is no reference to the Application class in that function

src/Magnum/ImGuiIntegration/Context.cpp
mosra: I see, that makes sense then :)

mosra:
> there is no reference to the Application class in that function

I think this could make it work -- add a new templated `Context` constructor overloads defined in `Context.hpp` that take a reference to the `Application`, delegate to the non-templated ones, and then set up additional flags based on what the application type supports.

```cpp
// Context.h
template<class Application> explicit Context(const Vector2& size, Application& application);

template<class Application> explicit Context(ImGuiContext& context, const Vector2& size, Application& application);
```

```cpp
// Context.hpp

// implicitly fetches framebuffer and window size from the application instance
template<class Application> Context::Context(const Vector2& size, Application& application): Context{size, application.windowSize(), application.framebufferSIze()} {
    if(application has warp cursor)
       ImGui::GetIO().BackendFlags |= ImGuiBackendFlags_HasSetMousePos;
}

template<class Application> Context::Context(ImGuiContext& context, const Vector2& size, Application& application): Context{context size, application.windowSize(), application.framebufferSIze()} {
    // same here, or maybe have some shared private helper for this instead?
}
```

The docs should then say something like "using this template constructor is preferrable as it enables additional features based on given Application capabilities". Not sure yet if the original constructor should be deprecated, probably not -- as I said before, I want to eventually split this code into an event handling and rendering part so they can be better mixed with 3rd party application toolkits or 3rd party renderers, so I'll do the deprecation as part of that instead.

src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp
```diff
@@ -114,8 +114,11 @@ struct Application {
         None = 999
     };
 
-    void setCursor(Cursor cursor) { currentCursor = cursor; }
-    Cursor currentCursor = Cursor::None;
+    Cursor _currentCursor = Cursor::None;
+    Vector2i _mousePos;
```
mosra, line 118:
```suggestion
    Cursor currentCursor = Cursor::None;
    Vector2i mousePos;
```

This could have stayed non-underscored, they're public variables.

mosra (commented):
Apart from that one trailing whitespace this looks good, now just the docs :)

mosra:
Merged as b376142148f17b3f33264b0ff17e78cffaf0ee7f.

PR #109, opened 2024-08-26
fix imageButton breaking change from imgui 5de7f69

kolbbond:
This recent imgui update broke the imageButton wrapper in "Widgets.h"

ocornut/imgui@5de7f69

I changed the arguments to fit the new imgui function.

Thanks.


pezcode:
I suppose this needs an ifdef around the imgui version define for compatibility with older versions.

I'll test this, maybe there's also a way to emulate the old behavior by generating an ID transparently. Then there's no need for breaking the function signature either.

mosra:
Merged as part of #110 in 541c850cd9f59c1efb5ada3e6a5de1e2cec735da, thank you both!

PR #110, opened 2024-09-15
ImGuiIntegration: fix imageButton() with 1.91.1 and up

pezcode:
Based on https://github.com/mosra/magnum-integration/pull/109, with added backwards compatibility.

I deprecated the old `imageButton()` function, not sure if that's a good idea, so feel free to drop that commit.

### DIFF
--- a/src/Magnum/ImGuiIntegration/Test/WidgetsGLTest.cpp
+++ b/src/Magnum/ImGuiIntegration/Test/WidgetsGLTest.cpp
@@ -107,7 +107,11 @@ void WidgetsGLTest::imageButton() {
 
     c.newFrame();
 
-    ImGuiIntegration::imageButton(texture, {100, 100});
+    ImGuiIntegration::imageButton("button", texture, {100, 100},
+        {{}, Vector2{1.0f}}, Color4::yellow(), Color4::blue());
+
+    ImGuiIntegration::imageButton(texture, {100, 100},
+        {{}, Vector2{1.0f}}, 5, Color4::yellow(), Color4::blue());
 
     c.drawFrame();
 
--- a/src/Magnum/ImGuiIntegration/Test/WidgetsGLTest.cpp
+++ b/src/Magnum/ImGuiIntegration/Test/WidgetsGLTest.cpp
@@ -110,8 +110,12 @@ void WidgetsGLTest::imageButton() {
     ImGuiIntegration::imageButton("button", texture, {100, 100},
         {{}, Vector2{1.0f}}, Color4::yellow(), Color4::blue());
 
+    #ifdef MAGNUM_BUILD_DEPRECATED
+    CORRADE_IGNORE_DEPRECATED_PUSH
     ImGuiIntegration::imageButton(texture, {100, 100},
         {{}, Vector2{1.0f}}, 5, Color4::yellow(), Color4::blue());
+    CORRADE_IGNORE_DEPRECATED_POP
+    #endif
 
     c.drawFrame();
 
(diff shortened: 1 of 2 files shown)


src/Magnum/ImGuiIntegration/Widgets.h
```diff
@@ -70,15 +104,33 @@ inline void image(GL::Texture2D& texture, const Vector2& size,
     padding
 @param backgroundColor  Background color, default @cpp 0x00000000_rgbaf @ce
 @param tintColor        Tint color, default @cpp 0xffffffff_rgbaf @ce
-@m_since_{integration,2019,10}
+@m_deprecated_since_latest Use @ref imageButton(const char*, GL::Texture2D&, const Vector2&, const Range2D&, const Color4&, const Color4&) instead.
 */
 inline bool imageButton(GL::Texture2D& texture, const Vector2& size,
```
mosra, line 109:
```suggestion
CORRADE_DEPRECATED("use imageButton(const char*, GL::Texture2D&, const Vector2&, const Range2D&, const Color4&, const Color4&) instead") inline bool imageButton(GL::Texture2D& texture, const Vector2& size,
```

src/Magnum/ImGuiIntegration/Widgets.h
```diff
@@ -70,15 +104,33 @@ inline void image(GL::Texture2D& texture, const Vector2& size,
     padding
 @param backgroundColor  Background color, default @cpp 0x00000000_rgbaf @ce
 @param tintColor        Tint color, default @cpp 0xffffffff_rgbaf @ce
```
mosra, line 106: I'd remove the `@param` lines here, as it's enough to have them in the non-deprecated API.

src/Magnum/ImGuiIntegration/Widgets.h
```diff
@@ -70,15 +104,33 @@ inline void image(GL::Texture2D& texture, const Vector2& size,
...
-@m_since_{integration,2019,10}
+@m_deprecated_since_latest Use @ref imageButton(const char*, GL::Texture2D&, const Vector2&, const Range2D&, const Color4&, const Color4&) instead.
```
mosra, line 107: I'd prefix this with `Using an implicit ID is no longer possible in newer ImGui.` or something similar, so it's clear why this got deprecated without having to look at the source.

src/Magnum/ImGuiIntegration/Test/WidgetsGLTest.cpp
```diff
@@ -107,7 +107,13 @@ void WidgetsGLTest::imageButton() {
 
     c.newFrame();
 
-    ImGuiIntegration::imageButton(texture, {100, 100});
+    ImGuiIntegration::imageButton("button", texture, {100, 100},
+        {{}, Vector2{1.0f}}, Color4::yellow(), Color4::blue());
+
+    #ifdef MAGNUM_BUILD_DEPRECATED
+    ImGuiIntegration::imageButton(texture, {100, 100},
```
mosra, line 114: ... and then you need to wrap this with `CORRADE_IGNORE_DEPRECATED_PUSH` and `_POP` to not warn when being compiled.

mosra (commented):
You're on a roll today! 

Deprecation is good, I'll keep that, it just needs a compiler warning as well. Other than that the code is perfect.

mosra:
Merged as f38a71a3082385f3ee82329ec0f7af016f95259b...df8739d05315e7663af83f06477f5788a8463cb2, thanks! (I amended the middle commit to add @kolbbond as a co-author, other than that it's unchanged.)

PR #111, opened 2024-09-17
ImGuiIntegration: bump minimum supported imgui version to 1.88

pezcode:
Bumping to 1.88, tagged on June 21st 2022 which is inline with [the two-year policy](https://github.com/mosra/magnum-integration/pull/98#issuecomment-1197986375).

Most of the removed code is preprocessor-branching for the obsolete pre-1.87 input handling.


mosra:
Thank you!

PR #112, opened 2024-10-14
Fix for ImGui breaking change: Use static cast instead of reinterpret cast for new ImTextureID

kolbbond:
Hello again.

Looks like another breaking imgui change from
https://github.com/ocornut/imgui/commit/92b94980c69ff3d3d7f5dc30c1c19abfb72db47e

Since the default ImTextureID is now Im64 instead of void* I believe we no longer need to reinterpret_cast as the type conversion could be done implicitly.
In any case I changed it to explicitly static_cast to the same types.
This compiles and works for my use-case.

Thanks.

### DIFF
--- a/src/Magnum/ImGuiIntegration/Context.cpp
+++ b/src/Magnum/ImGuiIntegration/Context.cpp
@@ -316,7 +316,14 @@ void Context::drawFrame() {
 
             /* We're storing just texture IDs, so make a non-owning instance
                around it, and assume it's already created */
-            GL::Texture2D texture = GL::Texture2D::wrap(static_cast<std::uintptr_t>(pcmd->TextureId), GL::ObjectFlag::Created);
+            GL::Texture2D texture = GL::Texture2D::wrap(
+                #if IMGUI_VERSION_NUM >= 19131
+                    pcmd->TextureId,
+                #else
+                    reinterpret_cast<std::uintptr_t>(pcmd->TextureId),
+                #endif
+                    GL::ObjectFlag::Created);
+
             _shader
                 .bindTexture(texture)
                 .draw(_mesh);
--- a/src/Magnum/ImGuiIntegration/Context.cpp
+++ b/src/Magnum/ImGuiIntegration/Context.cpp
@@ -316,7 +316,7 @@ void Context::drawFrame() {
 
             /* We're storing just texture IDs, so make a non-owning instance
                around it, and assume it's already created */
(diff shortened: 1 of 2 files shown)


pezcode:
Why doesn't the `reinterpret_cast` work anymore?

src/Magnum/ImGuiIntegration/Widgets.h
```diff
@@ -55,7 +55,7 @@ an implementation detail that might change in the future.
 @see @ref image(), @ref imageButton()
 */
 inline ImTextureID textureId(GL::Texture2D& texture) {
-    return reinterpret_cast<ImTextureID>(texture.id());
+    return static_cast<ImTextureID>(texture.id());
```
mosra, line 58: Could you make it so it still works on older versions? I suppose it'd be something like this -- and given that both types are now unsigned ints, it should work without any explicit cast:

```suggestion
    #if IMGUI_VERSION_NUM >= 19131
    return texture.id();
    #else
    return reinterpret_cast<ImTextureID>(texture.id());
    #endif
```

src/Magnum/ImGuiIntegration/Context.cpp
```diff
@@ -316,7 +316,7 @@ void Context::drawFrame() {
 
             /* We're storing just texture IDs, so make a non-owning instance
                around it, and assume it's already created */
-            GL::Texture2D texture = GL::Texture2D::wrap(reinterpret_cast<std::uintptr_t>(pcmd->TextureId), GL::ObjectFlag::Created);
+            GL::Texture2D texture = GL::Texture2D::wrap(static_cast<std::uintptr_t>(pcmd->TextureId), GL::ObjectFlag::Created);
```
mosra, line 319: And here like this -- I hope this works:

```suggestion
            GL::Texture2D texture = GL::Texture2D::wrap(
                #if IMGUI_VERSION_NUM >= 19131
                pcmd->TextureId,
                #else
                reinterpret_cast<std::uintptr_t>(pcmd->TextureId),
                #endif
                GL::ObjectFlag::Created);
```

mosra (commented):
Lol, that commit happened _one hour_ after I made the change in 1117d6f9b843aafe1fbd95bbadbc36e1180a488a (where I was sad that I had to use a `reinterpret_cast`). Suspiciously close, it's like if Omar was watching my changes here and decided to merge the PR right after.

Thanks for the PR! Two suggestions that should make it still build on older versions (and thus also the CI builds here), let me know if that works on your end as well.

mosra:
Awesome! I'm going to wait for the builds to sufficiently finish and will merge after.

mosra:
Merged as 92eb0d77a3372d7dff43b531c015b1bc0a339678, thank you!

PR #119, opened 2025-09-02
ImGuiIntegration: add support for dynamic textures

pezcode:
Adds support for dynamic texture creation and updating to the ImGui backend. This got added to ImGui in version 1.92 and mainly allows dynamic font atlasses for lower memory and font/DPI scaling without having to rebuild the whole atlas.

This sort of (but not really, see below) obsoletes https://github.com/mosra/magnum-integration/pull/115 because memory usage is now low enough for this to not matter.

ImGui still supports alpha-only textures, so it might be worth revisiting. For now only RGBA is supported, this matches what the official ImGui GL backend does.

## Todo

- [x] Tests
  - [x] Check that it compiles with 1.92.9 on CI
  - [x] Can re-use the text rendering test from https://github.com/mosra/magnum-integration/pull/115 
  - [x] Fix memory corruption due to mismatching defines of `IMGUI_DISABLE_OBSOLETE_FUNCTIONS`
  - [x] Clipboard context-awareness change results in random test failure (possibly fixed with the above, can't repro anymore)
  - [x] DPI scaling isn't working as it should, add a test
  - [x] Add support for single-channel texture format


src/Magnum/ImGuiIntegration/Context.cpp
```diff
@@ -48,12 +49,46 @@
...
+namespace {
+
+#if IMGUI_HAS_DYNAMIC_TEXTURES
+void updateTexture(ImTextureData& texture, const Range2Di& rect) {
+    const Vector2i offset = rect.min();
+    PixelStorage storage;
+    /* Data is tightly packed and rect can have any size */
+    storage.setAlignment(1);
+    storage.setRowLength(texture.Width);
+    storage.setSkip({offset.x(), offset.y(), 0});
```
pezcode, line 74: Is this safe on every platform, specifically WebGL1? The ImGui code does some manual packing into a tight buffer

src/Magnum/ImGuiIntegration/Context.cpp
```diff
@@ -9,7 +9,7 @@
     Copyright © 2018 Jonathan Hale <squareys@googlemail.com>
     Copyright © 2019 bowling-allie <allie.smith.epic@gmail.com>
     Copyright © 2021 Juan Pedro Bolívar Puente <raskolnikov@gnu.org>
-    Copyright © 2022, 2024 Pablo Escobar <mail@rvrs.in>
+    Copyright © 2022, 2025 Pablo Escobar <mail@rvrs.in>
```
mosra, line 12:
```suggestion
    Copyright © 2022, 2024, 2025 Pablo Escobar <mail@rvrs.in>
```

Same in the other cases.

src/Magnum/ImGuiIntegration/Context.cpp
```diff
@@ -271,6 +325,51 @@ void Context::drawFrame() {
...
+                    break;
+                }
+                case ImTextureStatus_WantDestroy:
+                    destroyTexture(*tex);
+                    /* destroyTexture() sets status and invalid ID */
+                    break;
+                case ImTextureStatus_WantUpdates:
+                    /* There's also the combined rect of all updates in
+                       tex->UpdateRect but the official GL backend doesn't use
+                       it either */
```
mosra, line 357: :eyes: 

I'd honestly prefer the combined rect here, especially given the overhead on WebGL (where it could be doing CPU-side copies and whatnot for each such call) in my opinion makes more sense to optimize for less GL calls than less memory copied.

With Vulkan there's IIRC a way to perform a copy of multiple rectangles in a single call so there it'd be preferred, but with GL I think it's easiest to just fire off one copy and hope for it to be DMA'd in some way.

src/Magnum/ImGuiIntegration/Context.hpp
```diff
@@ -498,13 +498,28 @@ namespace Implementation {
...
 
 /* Application implementations where returned clipboard contents don't have to
    be freed, such as GlfwApplication */
 template<class Application> struct ApplicationClipboard<Application, typename std::enable_if<std::is_same<decltype(std::declval<Application>().clipboardText()), Containers::StringView>::value>::type> {
-    static void connect(ImGuiIO& io, Context&, Application& application) {
+    static void connect(Context&, Application& application) {
+        #if IMGUI_VERSION_NUM >= 19110
+        ImGuiPlatformIO& platformIO = ImGui::GetPlatformIO();
+        platformIO.Platform_GetClipboardTextFn = [](ImGuiContext*) {
+            void* state = ImGui::GetPlatformIO().Platform_ClipboardUserData;
```
mosra, line 511: Can't you access the PlatformIO from the passed context pointer in some way?

I don't like how implicit the context is here, and looking closer I didn't have it correct originally either. It's called as `_imgui.connectApplicationClipboard(*this)` and one could originally assume that it deals with the context associated with `_imgui`, but it doesn't hold and it instead deals with whatever is "current". So, like, if I have `_imgui` and `_imgui2` in some test app, and call `connectApplicationClipboard()` on both, it'll do it just for one of the two and not both, or potentially some other context entirely, or nothing at all (crash?) if no ImGui context is current.

src/Magnum/ImGuiIntegration/Widgets.h
```diff
@@ -72,18 +72,40 @@ inline ImTextureID textureId(GL::Texture2D& texture) {
...
+{
+    ImGui::Image(textureId(texture), ImVec2(size), ImVec2(uvRange.topLeft()), ImVec2(uvRange.bottomRight()));
+}
+
+#ifdef MAGNUM_BUILD_DEPRECATED
+/**
+@brief Image widget displaying a @ref GL::Texture2D
+@m_deprecated_since_latest Using explicit tint and border color is no longer
+    possible in newer ImGui. Use @ref image(GL::Texture2D&, const Vector2&, const Range2D&)
+    instead.
```
mosra, line 89: Can you mention this in `doc/changelog-integration.dox`, potentially updating other deprecation notices for `image()`?

And isn't the same case with tint / border in `imageButton()`?

mosra (commented):
Regarding GLES test failures --- ah crap, I'm still relying on the shitty SwiftShader for the GL tests here, better not trust that pile of poo with anything. I'll look into updating the CI to use llvmpipe like elsewhere.

And yes I can use different imgui versions on different CI jobs, there's plenty of them. For example the GLES3 / non-deprecated builds could use the new versions, and desktop + GLES2 the oldest supported one. Don't want to mix it up too much tho, as with more variations it'd become hard to tell what exactly is failing, whether it's due to the OS, to the platform, to ImGui version or due to something else entirely.

src/Magnum/ImGuiIntegration/Context.cpp
pezcode: I copypasted what `GlyphCacheGL` does, kinda cleans up nicely when combined with only updating the combined rect

src/Magnum/ImGuiIntegration/Widgets.h
pezcode: Added. `ImGui::ImageButton()` still has the explicit tint and border color. I don't know why and I don't really want to know, either 🐱

pezcode: To be honest, I think all of these widget wrappers should be deprecated, seeing how much pain they've already caused. And with `textureId()` now being a thing, there's really no good reason to use these.

pezcode:
CircleCI seems to have issues 🌵 Some failure during `apt install` but I didn't touch any of that

pezcode:
I realized this is missing a change - the fonts are not rasterized with the DPI scaling. Don't merge yet please 🙇

mosra:
When you get back to this, be sure to rebase against current `master`, I switched from swiftshader to llvmpipe for better GL test coverage and adapted to various changes in the Windows CI runners that made the builds fail.

jvannugteren:
Hi pezcode, any news on this? I'm actually getting a segfault when when the font texture gets updated on a relayout. This happens for example when I turn the monitor off and on again. I assume this is related to "the fonts are not rasterized with the DPI scaling".

pezcode:
> Hi pezcode, any news on this? I'm actually getting a segfault when when the font texture gets updated on a relayout. This happens for example when I turn the monitor off and on again. I assume this is related to "the fonts are not rasterized with the DPI scaling".

Sorry, I kind of forgot about this PR. I'll try to wrap this up.

Re your segfault: I also occasionally saw this with the old font atlas but that was DPI scale somewhere along the way becoming 0. But I'll try your repro 🙇

jvannugteren:
Thanks for having a look. Some further information. Stacktrace reveals that the segfault occurs on `SetSubImage` on line 212 in the `relayout` function in MagnumIntegration/src/magnum/ImguiIntegration/Context.cpp (both with SDL2 and SDL2-compat). I wonder whether the font texture even needs updating anymore, as imgui seems to take care of it by itself nowadays, which might be the problem as magnum is interfering. Also the functions from imgui `ClearTexData` is now marked [OBSOLETE] and `SetTexID` is marked legacy.

src/Magnum/ImGuiIntegration/Context.cpp
```diff
@@ -48,12 +49,91 @@
...
+       support it either */
+    CORRADE_INTERNAL_ASSERT(texture.Format == ImTextureFormat_RGBA32);
+    CORRADE_INTERNAL_ASSERT(texture.GetTexID() == ImTextureID_Invalid);
+    const Vector2i size{texture.Width, texture.Height};
+    GL::Texture2D glTexture;
+    glTexture
+        .setMinificationFilter(SamplerFilter::Linear)
+        .setMagnificationFilter(SamplerFilter::Linear)
+        .setWrapping(GL::SamplerWrapping::ClampToEdge)
+        .setStorage(1, GL::TextureFormat::RGBA8, size);
```
pezcode: Todo: Needs special handling for GLES2, see old code

src/Magnum/ImGuiIntegration/Test/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 #               2020, 2021, 2022, 2023, 2024, 2025, 2026
 #             Vladimír Vondruš <mosra@centrum.cz>
 #   Copyright © 2018 Jonathan Hale <squareys@googlemail.com>
-#   Copyright © 2022 Pablo Escobar <mail@rvrs.in>
+#   Copyright © 2022, 2025 Pablo Escobar <mail@rvrs.in>
 #
```
pezcode, line 9: Todo: Bump these to 2026

src/Magnum/ImGuiIntegration/Test/CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ if(MAGNUM_BUILD_GL_TESTS)
             ContextTestFiles/draw.png
             ContextTestFiles/draw-scissor.png
             ContextTestFiles/draw-texture.png
+            ContextTestFiles/draw-text.png
```
pezcode, line 66: Todo: Add `draw-text-dpi-scaled.png` here as well

pezcode: This is causing the [ci/circleci: android-x86-gles2](https://circleci.com/gh/mosra/magnum-integration/3592) failure. Not sure why the Emscripten tests passed, investigate.

src/Magnum/ImGuiIntegration/Context.h
```diff
@@ -452,6 +458,11 @@ class MAGNUM_IMGUIINTEGRATION_EXPORT Context {
         /**
          * @brief Font texture used in `ImFontAtlas`
          * @m_since_{integration,2020,06}
+         *
+         * Returns the underlying texture for the ImGui font atlas. ImGui
+         * versions 1.92 and later create and destroy textures dynamically, on
+         * those versions this function returns an empty, default-constructed
+         * texture.
          */
         GL::Texture2D& atlasTexture() { return _texture; }
```
pezcode, line 510: Should probably be deprecated once versions before 1.92 are not supported anymore

src/Magnum/ImGuiIntegration/Test/CMakeLists.txt
mosra: This is because the Emscripten tests are not run inside a browser on the CI (too much headache involved). If you'd run them locally in a browser, it'd fail as well.

pezcode: Ah oke. Fixed this issue, but I'll make sure to run it locally as well.

package/ci/circleci.yml
```diff
@@ -636,7 +649,10 @@ jobs:
         version: "3.20.6"
         suffix: linux
         sudo: sudo
-    - install-imgui
+    # Test newer ImGui in at least one GLES2 job to cover the dynamic texture
+    # path in 1.92 and up
+    - install-imgui:
+        version: "v1.92.9"
```
mosra: I'd do this in the `linux-gles2` job instead, the Android jobs are running with a weird SwiftShader GL emulation that has a ton of bugs and it's better to have that code path tested on something reliable instead.

src/Magnum/ImGuiIntegration/Test/ContextGLTest.cpp
```diff
@@ -26,21 +26,13 @@
...
-   issues early. Doing this only in tests so the library itself can be used
-   with any newer version, but tests should be always run against the oldest
-   supported which is mentioned in doc/namespaces.dox, and which is downloaded
-   in all CI targets in package/ci/. The oldest supported version is tracked to
-   be roughly two years back. */
-#ifndef MAGNUM_BUILD_DEPRECATED
-#define IMGUI_DISABLE_OBSOLETE_FUNCTIONS
-#endif
-
+#include <cstring>
```
mosra:
```suggestion
#include <cstring> /* std::strcpy() */
```

src/Magnum/ImGuiIntegration/Context.cpp
```diff
@@ -48,12 +49,117 @@
...
+    const ImageView2D imageView{storage, pixelFormat, size, data};
+
+    GL::Texture2D glTexture = GL::Texture2D::wrap(GLuint(texture.GetTexID()), GL::ObjectFlag::Created);
+    glTexture.setSubImage(0, offset, imageView);
+    texture.SetStatus(ImTextureStatus_OK);
+}
+
+void destroyTexture(ImTextureData& texture) {
+    CORRADE_UNUSED GL::Texture2D glTexture = GL::Texture2D::wrap(GLuint(texture.GetTexID()),
+        GL::ObjectFlag::Created|GL::ObjectFlag::DeleteOnDestruction);
```
mosra: Instead of assigning to an unused variable and the UNUSED annotation, I'd call `wrap()` alone, ignore the return value, and add a comment about what is it doing.

src/Magnum/ImGuiIntegration/Context.cpp
```diff
@@ -48,12 +49,117 @@
...
+    /* On ES2 without EXT_unpack_subimage and on WebGL 1 there's no possibility
+       to upload just a slice of the input, upload the whole image instead */
+    Vector2i offset{NoInit};
+    Vector2i size{NoInit};
+    PixelStorage storage;
+    /* Data is tightly packed */
+    storage.setAlignment(1);
+    #ifdef MAGNUM_TARGET_GLES2
+    #ifndef MAGNUM_TARGET_WEBGL
+    if(!GL::Context::current().isExtensionSupported<GL::Extensions::EXT::unpack_subimage>())
```
mosra, line 124: In order to explicitly test this code path on the CI, you could do something similar to what I did for Magnum in mosra/magnum@d5df35599f0a65e94b5745ca4a64438305166db7. `GL_EXT_texture_rg` isn't being used here for anything I think, so it'd be just one addition to the relevant files in `package/`.

src/Magnum/ImGuiIntegration/Context.h
mosra, line 510: I wonder if this should be a `NoCreate`d instance instead, or maybe completely compiled out on 1.92+? It's an extra GL object that isn't used for anything by ImGui anymore, and if the user would attempt to read it or add anything to it, it silently wouldn't do anything, which is I think worse than getting an error at compile time that this isn't useful for anything anymore.

Hmm, or maybe keep it but mark as deprecated if building with ImGui 1.92+, and remove altogether for 1.92+ on non-deprecated builds?

mosra (commented):
Will go through this more thoroughly later, this is just the immediate things that I noticed. Thanks a lot for this!

src/Magnum/ImGuiIntegration/Context.h
pezcode: Actually it already is `NoCreate`d, the comment is incorrect 🙊  Your last suggestion probably makes the most sense, will do that.

src/Magnum/ImGuiIntegration/Test/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 #               2020, 2021, 2022, 2023, 2024, 2025, 2026
 #             Vladimír Vondruš <mosra@centrum.cz>
 #   Copyright © 2018 Jonathan Hale <squareys@googlemail.com>
-#   Copyright © 2022 Pablo Escobar <mail@rvrs.in>
+#   Copyright © 2022, 2026 Pablo Escobar <mail@rvrs.in>
```
mosra: Could keep 2025 in all these occurences as well, in addition to 2026 :)

That's the only remaining comment I have, went commit-by-commit with a fresh brain and everything looks great, first class work once again. Thank you.

src/Magnum/ImGuiIntegration/Context.h
pezcode: I implemented this, but the slightly ugly side product is that Context.h now needs to include imgui.h to check for the version (or more precisely, the `IMGUI_HAS_TEXTURES` macro) on non-deprecated versions.

mosra:
Thank you! :)

